PR #10, opened 2017-10-05
Crf Based Ner

aleksei-ai:
## Types of changes

## Checklist:


aleksei-ai:
Hi Saif,

1. By my opinion we should split project into at two modules: lib and
benchmarks.
I don't want to do it in this branch. Until it I think test is the best
place for benchmarks.

2.  Idea has too much suggestions.  Please provide details if it's
important.

3.  Why do you end it with "TestSpec" ? For example here there is not
"test" word http://www.scalatest.org/user_guide/selecting_a_style.

4. Done.

5. As I told it's a tech debt and it's waiting refactoring. I think it's
better because we had bad smelling code in our annotators and I didn't want
to continue and didn't want to finish refactoring to have problems with
merge. Now I want to keep it in this file at this branch and do refactoring
in separate branch.

Actually I can do it with point 1.

6.  Done.

7.
- Why are you storing something with key annotatorType? I'm reading this
code and I'm expecting to see something related to annotator but in our lib
I see metadata.annotatorType = .... result of annotator work
- I'm planning to finish refactoring (see 6) with Annotataed.scala. Having
pack and unpack will make it more consistent.

8. Partially Done. I've added some comments.

9. True, fixed.

10. Done

saifjsl:
Thank you Aleksei, here responses:

1. Sounds good, but I feel unsure into releasing those tests as they are in public. We can do that for now but should try to put it properly in place. What do you think of a different private repo for those tests for now? Also, a small concern is that there is no test inside the ner.crf package, shouldn't we have at least a dummy ner crf small test at least to make sure it works for unit test general purposes?

2. The things that mostly matter to me are:

- Not use **new** on case classes
- Add explicit variable type where it can help the reader / code review understand the flow
- Avoid meaningless variable names, such as a, b, c, i where it can help the reader understand the meaning
Less importantly:
- Replace _(i <- 0 until a.size)_ with _(i <- a.indices)_
- Use length not size in arrays and lists

3. I don't really know why, it was a standard since the beginning and I followed it. If there is a better proposal we should change all accordingly. For now it is split between TestSpec and Behaviors

5. Fair enough. But please help not forgetting about this!

7. I know it sounds arbitrary. But using annotatorType as a key for results, help the Finisher and other validation scripts access the results in a standard fashion. There are few exceptions such as the POS which results in key-> and tag-> keys. But in your NER case, if it applies for a single result per metadata, lets try to use annotatorType as key. (take a look at Annotation.flatten for instance). Since NER has a single result in metadata, I encourage to use annotatorType as key to make smooth Finisher output.

Only items 2 and 7 hold me back for merging in my opinion, what do you think?

Thanks
Saif

aleksei-ai:
2.

> The things that mostly matter to me are:

Not use new on case classes
> Add explicit variable type where it can help the reader / code review
> understand the flow

Less importantly:
>
Replace (i <- 0 until a.size) with (i <- a.indices)
> Use length not size in arrays and lists


I studied several StyleGuide's: twitter
<http://twitter.github.io/effectivescala/> and  standard
<http://docs.scala-lang.org/style/>. I haven't found this rules.
I'm ok to fix some of them in case I see it. I don't understand why are you
focusing on it?

Avoid meaningless variable names, such as a, b, c, i where it can help the
> reader understand the meaning
>

Please provide more details. I'm using a, b, c, and i in obvious places.
Like add(a: Int, b: Int). Or i for iteration.
Actually the same style used in StyleGuides.


7.

> I know it sounds arbitrary. But using annotatorType as a key for results,
> help the Finisher and other validation scripts access the results in a
> standard fashion. There are few exceptions such as the POS which results in
> key-> and tag-> keys. But in your NER case, if it applies for a single
> result per metadata, lets try to use annotatorType as key. (take a look at
> Annotation.flatten for instance). Since NER has a single result in
> metadata, I encourage to use annotatorType as key to make smooth Finisher
> output.


I used "tag" because NER has the exactly the same structure as POS. I even
unpack and pack it reusing the same logic.

If you require every annotator to store something using the same key
annotatorType it's better to create field in Annotation.

PR #42, opened 2017-10-30
Fix source stream miss first line

saifjsl:
Fixes resource helper not reading first lines from text files
## Description

ResourceHelper's touch to see if file exists, executes an unwanted readline. This is restarted once the file is confirmed to exist now.

## Types of changes

## Checklist:

- [] I have read the [CONTRIBUTING](http://nlp.johnsnowlabs.com/contribute.html) page.


src/main/scala/com/johnsnowlabs/nlp/util/io/ResourceHelper.scala
```diff
@@ -29,7 +29,8 @@ object ResourceHelper {
     val pipe: Option[InputStream] = try {
       val touchSource = getClass.getResourceAsStream(resource)
       Source.fromInputStream(touchSource)("UTF-8").getLines().take(1)
```
showy: Is not easier just to check that touchSource != null ?

saifjsl: Thanks, I tried but in some cases we may get empty-iterator which is not useful either. I'll review it again but I think that won't help much.

PR #84, opened 2018-01-12
Make lowercasing by Normalizer optional

lambdaofgod:
## Description
I've noted that Normalizer lowercases all tokens that are fed into it. I added `lowercase` parameter that, when set to `false`, doesn't lowercase tokens. By default it's `true` so it behaves like previous version.

## Motivation and Context
It seems it could be advantageous to not lowercase some inputs (for example if someone wanted to feed output of SparkNLP pipeline to some other algorithms that are case-sensitive).

## How Has This Been Tested?
I've checked old tests for Normalizer and they still work, and I added tests for `lowercase = true`.

## Types of changes

## Checklist:

### DIFF
--- a/docs/components.html
+++ b/docs/components.html
@@ -243,6 +243,7 @@ documentAssembler = new DocumentAssembler()
                                             <ul>
                                               <li>
                                                 setPattern(pattern): Regular expression for normalization, defaults [^A-Za-z]
+                                                setLowercase(value): lowercase tokens, default true
                                               </li>
                                             </ul>
                                                 <b>Example:</b><br>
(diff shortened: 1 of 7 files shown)


lambdaofgod:
Should I make this PR to `master` or `tokenizer-normalizer-params`? I thought about the latter, but since it wasn't updated for a bit it looked like I'd change over 100 files

saifjsl:
Looks good. I will test it shortly. Thank you for your contribution. Looking for more!

saifjsl:
I am thinking of how else the Normalizer could be improved... maybe standardizing dates? (perhaps using date matcher), clearing up some abbreviations (Sentence detector has some regex work with abbreviations) or something else..

Issue #121, opened 2018-02-26
Wrong FS when loading PerceptronModel


avenka11:
## Description
Saving a trained POS model to s3 and loading it back throws a `java.lang.IllegalArgumentException` exception with the message "Wrong FS, ... expected: hdfs...."

## Expected Behavior
A saved model (either standalone or as part of a spark PipelineModel) should be loadable from s3

## Current Behavior
While the model can be saved at the moment, reading it back either as a standalone model or as part of a `PipelineModel` throws an exception. A sample stacktrace looks like this:
```
java.lang.IllegalArgumentException: Wrong FS: s3://<bucket-name>/pos-models/anc-pos/fields/POS Model, expected: hdfs://<ip-address>:9000
  at org.apache.hadoop.fs.FileSystem.checkPath(FileSystem.java:651)
  at org.apache.hadoop.hdfs.DistributedFileSystem.getPathName(DistributedFileSystem.java:193)
  at org.apache.hadoop.hdfs.DistributedFileSystem.access$000(DistributedFileSystem.java:105)
  at org.apache.hadoop.hdfs.DistributedFileSystem$18.doCall(DistributedFileSystem.java:1118)
  at org.apache.hadoop.hdfs.DistributedFileSystem$18.doCall(DistributedFileSystem.java:1114)
  at org.apache.hadoop.fs.FileSystemLinkResolver.resolve(FileSystemLinkResolver.java:81)
  at org.apache.hadoop.hdfs.DistributedFileSystem.getFileStatus(DistributedFileSystem.java:1114)
  at org.apache.hadoop.fs.FileSystem.exists(FileSystem.java:1415)
  at com.johnsnowlabs.nlp.serialization.StructFeature.deserializeObject(Feature.scala:111)
  at com.johnsnowlabs.nlp.serialization.Feature.deserialize(Feature.scala:44)
  at com.johnsnowlabs.nlp.FeaturesReader$$anonfun$load$1.apply(ParamsAndFeaturesReadable.scala:13)
  at com.johnsnowlabs.nlp.FeaturesReader$$anonfun$load$1.apply(ParamsAndFeaturesReadable.scala:12)
  at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
  at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
  at com.johnsnowlabs.nlp.FeaturesReader.load(ParamsAndFeaturesReadable.scala:12)
[...]

PR #135, opened 2018-03-09
passing uri to retrieve the correct FileSystem

albertoandreottiATgmail:
On com.johnsnowlabs.nlp.serialization.Feature a FileSystem object is created, and used to verify that the provided path exists, if we don't pass the path at creation of the object the default filesystem is used.
We want the filesystem of the path to be used, instead.
## Motivation and Context

Fixes issue #121 

## How Has This Been Tested?

I created a script that writes a model to S3, and then reads it back from the same location.

Ubuntu.

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ x] Code improvements with no or little impact

## Checklist:

### DIFF
--- a/src/main/scala/com/johnsnowlabs/nlp/HasWordEmbeddings.scala
+++ b/src/main/scala/com/johnsnowlabs/nlp/HasWordEmbeddings.scala
@@ -50,7 +50,8 @@ trait HasWordEmbeddings extends AutoCloseable with ParamsAndFeaturesWritable {
 
 
   def deserializeEmbeddings(path: String, spark: SparkContext): Unit = {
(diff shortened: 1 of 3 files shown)


saifjsl:
Looks great! Adding it to the other places where we load filesystem from hadoopConfiguration

algoscale1:
I think it needs to be done while loading the embeddings here:
https://github.com/JohnSnowLabs/spark-nlp/blob/1.6.0/src/main/scala/com/johnsnowlabs/nlp/embeddings/SparkWordEmbeddings.scala#L68

saifjsl:
https://github.com/JohnSnowLabs/spark-nlp/pull/249

PR #165, opened 2018-04-02
Added option for sentence detection

atomobianco:
Added option for sentence detection to just use custom bounds

## Description
It was impossible to use just a set of custom bounds for sentence splitting.

Added a spark option to specify if only custom bound characters have to be considered or added to the existing methodology of splitting sentences.

## Motivation and Context
It was not possible to only use custom bounds without using also those of pragmatic segmenter. Now, an option can be used to force to use only the custom bounds supplied. 

## How Has This Been Tested?
Simply running existing tests.

## Types of changes

## Checklist:


saifjsl:
Thank you @atomobianco for contributing and fixing this bug.

I am wondering whether we should always do useOnlyCustom bounds when customBounds has been provided, since I am not sure it makes sense to provide customBounds and then using the standard bounds as well..

Maybe there is a use case where user wants to combine custom bounds and normal detection? Why would he provide custom bounds if he also wants standard bounds?

atomobianco:
Forcing the user to chose between custom bounds and pragmatic ones would certainly simplify things, also the _setSomething_ would suggest output to be more coherent with its name. 

Although I think that it might occur a use case in which one finds good segmentations with the default pragmatic rules, but wants to add one more regex to consider.

We can go with the exclusive `setCustomBounds` version as you suggest and if any issue request comes up with the aforementioned need, add a `addCustomBounds` later.

Would that be ok?

saifjsl:
After further analyzing it, we currently have many use cases where custom bounds are mixed in along default bounds.

The most common use case I've seen is end line separators where it is expected they are considered on text aside from the usual sentence bounds.

So I'll keep your plan. I am about to commit a few changes and include it on next release

atomobianco:
Awesome. Thanks for your help improving the PR!

Issue #92, opened 2018-01-25
DateMatcher should create RuleFactory only once


saifjsl:
## Description

DateMatcher is creating a RuleFactory on every attempt to match a date. Should be created once right before the annotate() call if possible

PR #174, opened 2018-04-15
DateMatcher create regex factories only once

saifjsl:
## Description

Date Matcher is slightly faster by creating Rule Factories only once


saifjsl:
https://github.com/JohnSnowLabs/spark-nlp/issues/92

PR #229, opened 2018-06-17
Updated tensorflow structure, and spark default builder

saifjsl:
## Description

Enable tensorflow allow_soft_placement for gpu based models support. Updates TF dependency.

## Types of changes

## Checklist:

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,3 +1,5 @@
+import sbtassembly.MergeStrategy
+
 val sparkVer = "2.3.0"
 val scalaVer = "2.11.12"
 val scalaTestVersion = "3.0.0"
@@ -9,14 +11,14 @@ organization := "com.johnsnowlabs.nlp"
 
 version := "1.5.4"
 
-scalaVersion := scalaVer
+scalaVersion in ThisBuild := scalaVer
 
-sparkVersion := sparkVer
+sparkVersion in ThisBuild := sparkVer
 
 /** Spark-Package attributes */
-spName := "JohnSnowLabs/spark-nlp"
+spName in ThisBuild := "JohnSnowLabs/spark-nlp"
 
-sparkComponents ++= Seq("mllib")
+sparkComponents in ThisBuild ++= Seq("mllib")
 
 licenses += "Apache-2.0" -> url("http://opensource.org/licenses/Apache-2.0")
 
@@ -24,7 +26,7 @@ spIncludeMaven := false
 
 spAppendScalaVersion := false
 
-resolvers += "Maven Central" at "http://central.maven.org/maven2/"
+resolvers in ThisBuild += "Maven Central" at "http://central.maven.org/maven2/"
 
 assemblyOption in assembly := (assemblyOption in assembly).value.copy(
   includeScala = false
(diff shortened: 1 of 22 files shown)


build.sbt
```diff
@@ -88,7 +88,16 @@ lazy val utilDependencies = Seq(
   "org.slf4j" % "slf4j-api" % "1.7.25",
   "org.apache.commons" % "commons-compress" % "1.15",
   "org.tensorflow" % "tensorflow" % "1.8.0",
-  "com.amazonaws" % "aws-java-sdk-s3" % "1.11.313"
+  // Enable the following for tensorflow support
+  //"org.tensorflow" % "libtensorflow" % "1.8.0",
```
albertoandreottiATgmail: How did we manage to get rid of TF dependency here? Where are we getting that dependency from now?

build.sbt
```diff
@@ -88,7 +88,16 @@ lazy val utilDependencies = Seq(
...
+  //"org.tensorflow" % "libtensorflow_jni_gpu" % "1.8.0",
+  "com.amazonaws" % "aws-java-sdk-s3" % "1.11.313" exclude("org.slf4j", "slf4j-log4j12") exclude("log4j", "log4j"),
+  "org.apache.pdfbox" % "pdfbox" % "2.0.9" exclude("org.slf4j", "slf4j-log4j12") exclude("log4j", "log4j"),
+  "net.sourceforge.tess4j" % "tess4j" % "4.0.2" exclude("org.slf4j", "slf4j-log4j12") exclude("log4j", "log4j"),
```
albertoandreottiATgmail: All these excludes can be replaced with only one exclude on Tesseract dependency, like this,
"net.sourceforge.tess4j" % "tess4j" % "4.0.2" exclude("org.slf4j", "slf4j-log4j12-1.7.")

src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorflowWrapper.scala
```diff
@@ -96,14 +96,28 @@ object TensorflowWrapper {
...
     // 3. Read file as SavedModelBundle
-    val graphDef = Files.readAllBytes(Paths.get(folder, "saved_model.pb"))
-    val graph = new Graph()
-    graph.importGraphDef(graphDef)
-    val session = new Session(graph)
-    session.runner.addTarget("save/restore_all")
-      .feed("save/Const", t.createTensor(Paths.get(folder, "variables").toString))
-      .run()
+    val (graph, session) = if (useBundle) {
+      val model = SavedModelBundle.load(folder, tags:_*)
```
albertoandreottiATgmail: Curiosity question: any particular advantage to using this? best integration with Python maybe?

src/main/scala/com/johnsnowlabs/nlp/HasWordEmbeddings.scala
```diff
@@ -21,16 +21,20 @@ trait HasWordEmbeddings extends AutoCloseable with ParamsAndFeaturesWritable {
 
   val nDims = new IntParam(this, "nDims", "Number of embedding dimensions")
   val indexPath = new Param[String](this, "indexPath", "File that stores Index")
+  val normalizeEmbeddings = new BooleanParam(this, "normalizeEmbeddings", "whether to use embeddings of normalized tokens (if not already normalized)")
```
albertoandreottiATgmail: That parameter name brings to me the idea of normalizing the vectors(L2 norm, and the like), and not the tokens. Maybe normalizeTokens could work better?

build.sbt
saifjsl: I didn't, it's two lines above. The libtensorflow and libtensorflow-gpu are required for gpu usage

build.sbt
saifjsl: I don't remember making this change, but okay

src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorflowNer.scala
```diff
@@ -51,7 +49,7 @@ class TensorflowNer
         .feed(wordLengthsKey, tensors.createTensor(batchInput.wordLengths))
         .feed(charIdsKey, tensors.createTensor(batchInput.charIds))
 
-        .feed(dropoutKey, tensors.createTensor(1.1f))
+        .feed(dropoutKey, tensors.createTensor(1.0f))
```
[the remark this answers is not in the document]
saifjsl: I don't have much to say here. We do use the dropout value when validating and doing cross-validation during training. I belive this is making it up with more epochs. What do you suggest?

src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorflowWrapper.scala
saifjsl: Pros: Use Tensorflow API for loading model
Cons: There is no Save API for ModelBundle in JAVA, so we cannot train using this. Using SavedModelBundle is only useful for reading complete models trained in python. Otherwise, we need to use Aleksei's decomposed loading method to be able to save it manually during training (using python provided graph)

src/main/scala/com/johnsnowlabs/nlp/HasWordEmbeddings.scala
saifjsl: This is a good point. Although we are not normalizing tokens, just normalizing the tokens for retrieveing embeddings. I'll think of something better.

PR #255, opened 2018-08-09
Fixing misspelled parameter in Lemmatizer example

maziyarpanahi:
This fixes the Scala example of Lemmatizer in the documentation.

## Types of changes
Documentation


saifjsl:
Thank you

PR #339, opened 2019-01-09
Fix SentenceDetector  bug when bound is between two words that contain apostrophes

### LINKED ISSUE #276: sentenceDetector fails when bound is between two words that contain apostrophes
## Description

`SentenceDetector` annotator fails to detect bounds when bound is between two words that contain apostrophes.

## Expected Behavior
In such case, it should split such text based on the given bounds.
For example: 
> "We don't want to ignore this period. Isn't it right?"

Should become:
> "We don't want to ignore this period."
    "Isn't it right?"

## Current Behavior
It doesn't split text on such cases.
For example, the sentence: 
> "We don't want to ignore this period. Isn't it right?"

Stays the same.

## Possible Solution
The reason that it fails, is because it tries to protect everything within single quotes from being splitted. 
My (simple) suggestion is to consider two apostrophes as single quotes  only if the first one is coming at the beginning of the word. 

Practically I suggest to replace the following line of code from

https://github.com/JohnSnowLabs/spark-nlp/blob/c2371f58f7d91221bbfa9265debd3afd10588c90/src/main/scala/com/johnsnowlabs/nlp/annotators/sbd/pragmatic/PragmaticContentFormatter.scala#L326

to:
`.addRule(new RegexRule("(?<=\\s|^)'(?:[^']|'\\w)+'", "betweenPunctuations-singleQuot"))`

haimco10:
## Description

Here I added a test in 'PragmaticApproachTestSpec' to show that current single quot protection version fails. Then I fixed the single quotes regex. With this fix it passes all the the previous tests and the new one.   

## Motivation and Context

This PR should resolve #276 . 

In case when we have sentence bound (e.g. period) between two words that contain apostrophes, the `SentenceDetector` failed to split the multiple sentences and leave it as one long sentence. 
It happens because the it tries to protect everything within the two single quotes from been broken to sentences. This is a good behavior, but there are cases where it apostrophes with single quotes. 

## How Has This Been Tested?

I wrote a new test that in `PragmaticApproachTestSpec`. This test shows that when we apply  the current version of `MixedPragmaticMethod` to the text: 

> We don't want to ignore this period. Isn't it right?

 fails to split it to two sentences. With this fix, it passes the test and split it into two sentences.
In addition it still passes the rest of the tests.

## Types of changes

## Checklist:


saifjsl:
This is looking great! Thank you! I am taking a look and so far so good as you made everyone happy

PR #390, opened 2019-02-19
Spark-nlp in 30 seconds

saifjsl:
## Description

Includes changes that make Spark-NLP easy for new comers, by calling `sparknlp.session()`. Also changes PretrainedPipeline api in order to easy retrieve pretrained pipelines.

## Types of changes

## Checklist:


python/example/model-downloader/ModelDownloaderExample.ipynb
```diff
@@ -198,9 +198,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from sparknlp.pretrained.pipeline.en import SentimentPipeline\n",
-    "\n",
-    "SentimentPipeline.annotate(\"This is a good movie!!!\")"
+    "PretrainedPipeline(\"pipeline_vivekn\").annotate(\"This is a good movie!!!\")"
```
maziyarpanahi: Shouldn't this be `sentiment_vivken`? We know already it is a pipeline due to `PretrainedPipeline`, I think `vivekn` on its own is not clear, so we can combine the annotator with its algorithm/method: `sentiment_vivekn`.

python/sparknlp/pretrained.py
```diff
@@ -0,0 +1,45 @@
...
+from sparknlp.annotator import *
+from pyspark.ml.wrapper import JavaModel
+from pyspark.sql import DataFrame
+from sparknlp.base import LightPipeline
+
+
+class ResourceDownloader(object):
+
+    @staticmethod
+    def downloadModel(reader, name, language, remote_loc=None):
```
maziyarpanahi: Can we also add another field called `source` which represents the source of the dataset which we built the model? For instance, we will have multiple Lemma in Italian one from DXC, one from UD, one from the famous GitHub repo for Lemma Dictionaries, etc.
This way, we can have `source` default to one of these models, and the user can choose whether they wan POS Tagger in English based on regular format or the one from the UD treebank.

src/main/scala/com/johnsnowlabs/nlp/pretrained/PretrainedPipeline.scala
```diff
@@ -0,0 +1,29 @@
+package com.johnsnowlabs.nlp.pretrained
+
+import com.johnsnowlabs.nlp.LightPipeline
+import org.apache.spark.ml.PipelineModel
+import org.apache.spark.sql.DataFrame
+
+case class PretrainedPipeline(
+                               downloadName: String,
+                               language: String = "en",
+                               source: String = ResourceDownloader.publicLoc
```
maziyarpanahi: Shoot! I can see we have a filed here called `source` which is about the location of the model. If we decide to add a field to allow users to choose which model they want for one specific annotator, maybe we can choose `type` or something or rename this to `url` and use the `source`.

src/test/scala/com/johnsnowlabs/nlp/annotators/ner/dl/NerDLReader.scala
```diff
@@ -28,13 +28,13 @@ class NerDLReaderTestSpec extends FlatSpec {
     val spark = ResourceHelper.spark
     import spark.implicits._
 
-    val bp = BasicPipeline().pretrained()
+    val bp = PretrainedPipeline("pipeline_basic")
```
maziyarpanahi: Should we just call this `basic` instead of `pipeline_basic`?

saifjsl: subject to renew, we should create a new list of pipelines we want to have. this specific test is using that one so we might keep it just fbecause

src/main/scala/com/johnsnowlabs/nlp/pretrained/PretrainedPipeline.scala
saifjsl: right, agreed. I'll rename that parameter

python/example/model-downloader/ModelDownloaderExample.ipynb
saifjsl: disregard model and pipeline names for now :) they should be all remade from scratch

python/sparknlp/pretrained.py
saifjsl: For now we could name it just differently, once we get to a point to multiple sources per public model we can re-consider

PR #413, opened 2019-02-28
MaxLength now truncates in whitespaces, defaults to disabled

saifjsl:
## Types of changes

## Checklist:

### DIFF
--- a/src/main/scala/com/johnsnowlabs/nlp/annotators/sbd/SentenceDetectorParams.scala
+++ b/src/main/scala/com/johnsnowlabs/nlp/annotators/sbd/SentenceDetectorParams.scala
@@ -2,6 +2,8 @@ package com.johnsnowlabs.nlp.annotators.sbd
 
 import org.apache.spark.ml.param.{BooleanParam, IntParam, Params, StringArrayParam}
 
+import scala.collection.mutable.ArrayBuffer
+
 trait SentenceDetectorParams extends Params {
 
   val useAbbrevations = new BooleanParam(this, "useAbbreviations", "whether to apply abbreviations at sentence detection")
@@ -22,7 +24,6 @@ trait SentenceDetectorParams extends Params {
     useAbbrevations -> true,
     useCustomBoundsOnly -> false,
     explodeSentences -> false,
-    maxLength -> 240,
     customBounds -> Array.empty[String]
   )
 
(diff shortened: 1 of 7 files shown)


src/main/scala/com/johnsnowlabs/nlp/annotators/sbd/SentenceDetectorParams.scala
```diff
@@ -38,4 +39,31 @@ trait SentenceDetectorParams extends Params {
...
+    val currentSentence = ArrayBuffer.empty[String]
+
+    def addWordToSentence(word: String): Unit = {
+      /** Adds +1 because of the space joining words */
+      currentLength += word.length + 1
+      currentSentence.append(word)
+    }
+
+    sentence.split(" ").foreach(word => {
+      if (currentLength + word.length > maxLength) {
```
correlator: 🙌

src/main/scala/com/johnsnowlabs/nlp/annotators/sbd/SentenceDetectorParams.scala
```diff
@@ -38,4 +39,31 @@ trait SentenceDetectorParams extends Params {
...
   def getMaxLength: Int = $(maxLength)
 
+  def truncateSentence(sentence: String, maxLength: Int): Array[String] = {
+    var currentLength = 0
+    val allSentences = ArrayBuffer.empty[String]
+    val currentSentence = ArrayBuffer.empty[String]
+
+    def addWordToSentence(word: String): Unit = {
+      /** Adds +1 because of the space joining words */
+      currentLength += word.length + 1
```
correlator: Is there a reason to set the max length by number of characters instead of number of tokens?

maziyarpanahi: Limiting the length of sentence to a number of tokens is not a good idea. If you fail to tokenize correctly like tokenizing the entire URL as one token, if you have multiple URLs, you will end up having 10 tokens in a sentence each around 50 characters (500 characters). Limiting the length by the number of tokens defeats the purpose of avoiding large sentences. 

Really hard to set a correct size to limit if it was the number of tokens. (limit by number of characters is the best practice same as in CoreNLP)

correlator: That's an interesting point about tokenization errors that result from missing spaces between words, or in long URLS.  However, this addition that requires splitting on whitespace will prevent this solution from solving the problem you just outlined @maziyarpanahi .  This feature may be worth thinking about in more detail.

I believe this feature was created to address another issue around a dl model having a max length setting that was throwing errors when embedded sentences were too long (too many tokens), an issue which this pr would theoretically solve.

I think that is why the maxLength setting is turned off by default.

saifjsl: Annotators should be capable of handling their own limits if needed. This change is towards that a little bit. This sentence limit might help universal problems or limitations or very specific use cases, but ideally, annotator specific limitations should be addressed locally.

Agreed on the points pointed out above!

PR #416, opened 2019-03-01
Create POS class to annotate a dataset and train Perceptron from DataFrame

maziyarpanahi:
## Description

Changing the way `Perceptron` is being trained by DataFrame in `.fit()` stage:

- [x] Create `POS` class and `readDataset` function to read text files and create DataFrame with column: `annotatorType.POS`
- [x] Change `PerceptronApproch` reading from `posCol` annotated column in `pos` style inside the DataFrame
- [x] Update Perceptron test with `posCol` to train from DataFrame generated by `POS().readDataset()`
- [x] Remove `setCorpus` and `ExternalResources`
- [x] Update all unit tests using `PerceptronApproach` the old way by using `setCorpus`

## Motivation and Context

Deprecating `.setCorpus` in training `PerceptronModel` and only use `.fit()` which is Spark ML style. `fit()` now uses a column (annotatorType.POS) in DataFrame that was generated from `POS().readDataset` class.

## Types of changes

## Checklist:


src/main/scala/com/johnsnowlabs/nlp/annotators/pos/perceptron/PerceptronApproach.scala
```diff
@@ -89,26 +76,26 @@ class PerceptronApproach(override val uid: String) extends AnnotatorApproach[Per
...
-      dataset.select(tokenColumn, $(posCol))
-        .as[(Array[Annotation], Array[String])]
+
+      val datasetSchemaFields = dataset.schema.fields
+        .find(f => f.metadata.contains("annotatorType") && f.metadata.getString("annotatorType") == AnnotatorType.POS)
+
+      require(datasetSchemaFields.map(_.name).isDefined, s"Cannot train from DataFrame without POS annotatorType by posCol")
+
+      val posColumn = datasetSchemaFields.map(_.name).get
+
```
maziyarpanahi: Small note here:
```
val datasetSchemaFields = dataset.schema.fields
        .find(f => f.metadata.contains("annotatorType") && f.metadata.getString("annotatorType") == AnnotatorType.POS
```
This will result either in the name of column with `AnnotatorType.POS` or in `None`.  The reason I seprate the `map(_.name).get` and put it after `require` was to make sure it doesn't crash in case there is no `POS` column without telling user why it crashed. 
The `require` makes sure user knows there is no `POS` clolumn instead of crashing without explaination.

src/main/scala/com/johnsnowlabs/nlp/datasets/POS.scala
```diff
@@ -0,0 +1,114 @@
...
+      .drop("tokens", "token_tags")
+
+    tempDataFrame.withColumn(
+      outPutPosColName,
+      wrapColumnMetadata(tempDataFrame(outPutPosColName), annotatorType, outPutPosColName)
+    )
+  }
+
+  // For testing purposes when there is an array of tokens and an array of labels
+  def readDatframe(posDataframe: DataFrame, tokensCol: String = "tokens", labelsCol: String = "labels",
```
maziyarpanahi: I create this function for our internal testing where we have an array of tokens and an array of labels instead of a path to text files.

saifjsl: Wondering how to make the import of this functions more intuitive. Maybe `import com.johnsnowlabs.nlp.training` ?

PR #589, opened 2019-08-12
WIP: Sentencepiece

alexander-n-thomas:
I've added the ability to call sentencepiece from the JVM by incorporating the code from the https://github.com/levyfan/sentencepiece-jni repo.

Branched from the quality_improvements branch, so it should follow that one, though there is no actual dependency.

## Description

The java code expects to have a native library on the classpath, or provided via sentencepiece.lib.path property. If these are not present, there will be a linking error. I've added unit tests (converted from levyfan repo). The tests only run if the native library is available

## Motivation and Context

We want to support XLNet which requires sentencepiece. The sentencepiece library is a python wrapper on a very large C++ library. This is the least evil fast implementation.

## How Has This Been Tested?

I have run the tests from the original repo. I have also tested to make sure that the unit tests do not run without the native library

## Types of changes

## Checklist:


saif-ellafi:
Closed for lack of activity

PR #619, opened 2019-09-06
Py embeddings format arg + fixes to Tokenizer and DateMatcher

sllynn:
## Description

Changes to python API only.

*WordEmbeddings*
Replaced the if/then/else block for selecting `format` argument with case insensitive Enum check.
Catch for failure with informative error msg.

*Tokenizer*
Added default values for some args.
Class now passes `ParamsGettersTestSpec` unit test.

*DateMatcher*
Appended `toString` method to `TypeConverter`.
Class now passes `DateMatcherTestSpec` unit test.

## Motivation and Context

Fixed a mismatch between documented options for `format` argument to `WordEmbeddings` annotator and python API (which expected all caps).

## How Has This Been Tested?

Ran unit tests (excl. `ContextSpellCheckerTestSpec`). Existing tests cover changes.
Pyspark 2.4.3, Spark-NLP 2.2.0 built from master branch.

## Types of changes

## Checklist:


python/sparknlp/annotator.py
```diff
@@ -99,7 +100,12 @@ def __init__(self):
         self._setDefault(
             targetPattern="\\S+",
             contextChars=[".", ",", ";", ":", "!", "?", "*", "-", "(", ")", "\"", "'"],
-            caseSensitiveExceptions=True
+            caseSensitiveExceptions=True,
+            exceptions=[],
+            infixPatterns=[],
+            prefixPattern="\\A([^\\s\\w\$\\.]*)",
+            splitChars=[],
+            suffixPattern="([^\\s\\w]?)([^\\s\\w]*)\\z')"
```
saifjsl: These defaults do not match the defaults on the Scala side. Also empty is not equal to unset, which is a subtle but important difference. The code sometimes check if the user defined any value at all, rather than an empty content.

python/sparknlp/annotator.py
```diff
@@ -470,7 +476,7 @@ class DateMatcher(AnnotatorModel):
     dateFormat = Param(Params._dummy(),
                        "dateFormat",
                        "desired format for dates extracted",
-                       typeConverter=TypeConverters)
+                       typeConverter=TypeConverters.toString)
```
saifjsl: good catch!

python/sparknlp/annotator.py
```diff
@@ -1106,8 +1112,6 @@ class NerDLApproach(AnnotatorApproach, NerApproach):
     useContrib = Param(Params._dummy(), "useContrib", "whether to use contrib LSTM Cells. Not compatible with Windows. Might slightly improve accuracy.", TypeConverters.toBoolean)
     trainValidationProp = Param(Params._dummy(), "trainValidationProp", "Choose the proportion of training dataset to be validated against the model on each Epoch. The value should be between 0.0 and 1.0 and by default it is 0.0 and off.",
                                 TypeConverters.toFloat)
-    includeValidationProp = Param(Params._dummy(), "includeValidationProp", "Whether or not to include trainValidationProp inside training or keep it for real sampling evaluation.",
-                                  TypeConverters.toBoolean)
```
saifjsl: why are you removing this?

python/sparknlp/annotator.py
```diff
@@ -1160,9 +1164,6 @@ def setTrainValidationProp(self, v):
         self._set(trainValidationProp=v)
         return self
 
-    def setIncludeValidationProp(self, v):
-        return self._set(includeValidationProp=v)
-
```
saifjsl: Still not sure why is the removal of this Param?

python/sparknlp/annotator.py
```diff
@@ -1523,20 +1523,20 @@ def __init__(self):
...
-        elif frmt == "TEXT":
-            return 2
-        elif frmt == "BINARY":
-            return 3
-        else:
-            return frmt
+    class Format(Enum):
+        SPARKNLP = 1
+        TEXT = 2
+        BINARY = 3
```
saifjsl: Thanks, this is much better

python/sparknlp/annotator.py
sllynn: I don't think I did, perhaps it was lost in the rebase?

python/sparknlp/annotator.py
sllynn: Fair enough, I was trying to insert sane defaults based on the docstring.

python/sparknlp/annotator.py
sllynn: As above, I can bring these in.

python/sparknlp/annotator.py
saifjsl: could be, merging with latest master should fix it

PR #635, opened 2019-09-13
setting the max buffer limit for kyroserializer

vkocaman:
setting the max buffer limit for kyroserializer while initializing the spark with sparknlp

## Description
max buffer set to 600M to get over kyroserializer error while integrating SparkML transformers within SparkNLP annotaters in the same pipeline

## Motivation and Context
when I tried to start the session with SparkSession  and set the configs properly (kyroserializer etc), I am not able to use sparknlp annotators.. for example, DocumentAssembler() is not callable.. but when I start the session with sparknlp.start(), it's fine.. However, in that case, I am not able to use sparkML transformers such as Word2Vec inside the pipeline with sparkNLP annotators.. it gives kyroserializer not enough buffer error.. so we may need to set max buffer while init Spark through SparkNLP.

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ x] Code improvements with no or little impact
- [ x] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/python/sparknlp/__init__.py
+++ b/python/sparknlp/__init__.py
@@ -36,7 +36,8 @@ def start(include_ocr=False):
         .appName("Spark NLP") \
         .master("local[*]") \
         .config("spark.driver.memory", "6G") \
-        .config("spark.serializer", "org.apache.spark.serializer.KryoSerializer", "600M")
+        .config("spark.serializer", "org.apache.spark.serializer.KryoSerializer") \
+        .config("spark.kryoserializer.buffer.max", "600M")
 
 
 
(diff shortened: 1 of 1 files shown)


maziyarpanahi:
This is a very use-case oriented situation. The `spark.kryoserializer.buffer.max` ranges from 64k to 2048m. Every use case requires a different value. The right way is to manually construct `SparkSession` with all the desire configs.

The combination of your Dataset and what you are doing in Spark NLP and Spark ML requires 600 (this time, this could change for other projects). It's not just Spark NLP + Spark ML, it's about your use case and your RDDs. (We have examples of Spark NLP and Spark ML especially for Wrd2Vec which is fine with the default value)

I suggest keeping `sparknlp.start()` clean with minimum configs as it's supposed to get the beginners up and running on a local machine. We advise advanced users such as you to have their own `SparkSession` either constructed locally or setup for a cluster.

PR #674, opened 2019-11-07
Optimize BertEmbeddings memory consumption

maziyarpanahi:
This PR intends to deal with `BertEmbeddings` annotator memory consumption.

- [x] Create tensors by using buffer instead of an object to improve performance:
reference: https://github.com/tensorflow/tensorflow/issues/8244#issuecomment-285459455

- [x] Clear sessions output to clear memory usage/consumption by `BertEmbeddings`

### DIFF
--- a/src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorResources.scala
+++ b/src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorResources.scala
@@ -25,7 +25,7 @@ class TensorResources {
     result
   }
 
-  def createBertTensor[T](shape: Array[Long], buf: IntBuffer): Tensor[_] = {
+  def createIntBufferTensor[T](shape: Array[Long], buf: IntBuffer): Tensor[_] = {
 
     val result = Tensor.create(shape, buf)
 
--- a/src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorResources.scala
+++ b/src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorResources.scala
@@ -2,8 +2,10 @@ package com.johnsnowlabs.ml.tensorflow
 
 import java.nio.{FloatBuffer, IntBuffer, LongBuffer}
 
+import com.fasterxml.jackson.annotation.JsonFormat.Shape
 import org.tensorflow.Tensor
 
+import scala.collection.mutable
 import scala.collection.mutable.ArrayBuffer
 import scala.language.existentials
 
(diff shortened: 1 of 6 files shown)


src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorResources.scala
```diff
@@ -12,24 +14,40 @@ class TensorResources {
...
+        Tensor.create(obj)
     }
 
     tensors.append(result)
     result
   }
 
+  def createIntBufferTensor[T](shape: Array[Long], buf: IntBuffer): Tensor[_] = {
+
+    val result = Tensor.create(shape, buf)
```
saif-ellafi: Is there any chance the Tensor we are creating is already on the buffer?

src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorResources.scala
```diff
@@ -12,24 +14,40 @@ class TensorResources {
...
 
   def createTensor[T](obj: T): Tensor[_] = {
-    val result = if (obj.isInstanceOf[String]) {
-      Tensor.create(obj.asInstanceOf[String].getBytes("UTF-8"), classOf[String])
-    }
-    else {
-      Tensor.create(obj)
+    val result = obj match {
+      case str: String =>
+        Tensor.create(str.getBytes("UTF-8"), classOf[String])
```
saif-ellafi: Is this a style improvement or is the actual casting to string expensive?

src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorflowBert.scala
```diff
@@ -21,36 +23,42 @@ class TensorflowBert(val tensorflow: TensorflowWrapper,
...
-      .run()
 
-    tensors.clearTensors()
+    val outs = runner.run().asScala
+    val embeddings = TensorResources.extractFloats(outs.head)
 
-    val embeddings = TensorResources.extractFloats(calculated.get(0))
+    tensors.clearSession(outs)
+    tensors.clearTensors()
+    buf.clear()
```
saif-ellafi: nice!!

src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorResources.scala
maziyarpanahi: No, the data we are sending to the previous `Tensor.create()` was object and TF decides the shape and the rest by itself. That is why it's less optimized and it is proffered to use `Tensor.create(shape, buffer)`.

src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorResources.scala
maziyarpanahi: It is recommended to use case instead of `obj.isInstanceOf[String]`

src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorflowBert.scala
maziyarpanahi: Yes, the `tensors.clearSession` is the part we get back all the memory! The improvement of going from +40G memory to stay at 8G over 100K sentences.

PR #681, opened 2019-11-14
Keep original index for the sentence-token columns pair

jfernandrezj:
## Description
Just avoiding the filtering of sentences with no tokens. Filtering sentences seems to have no use case at all, it actually makes everything from then on untraceable.

## Motivation and Context
To be consistent with sentence indexing all throught the annotation Pipeline

## How Has This Been Tested?
Same tests run

## Types of changes

## Checklist:

- [X] All new and existing tests passed.

### DIFF
--- a/src/main/scala/com/johnsnowlabs/nlp/annotators/common/TokenizedWithSentence.scala
+++ b/src/main/scala/com/johnsnowlabs/nlp/annotators/common/TokenizedWithSentence.scala
@@ -24,8 +24,7 @@ object TokenizedWithSentence extends Annotated[TokenizedSentence] {
         token.begin >= sentence.start & token.end <= sentence.end
       ).map(token => IndexedToken(token.result, token.begin, token.end))
       sentenceTokens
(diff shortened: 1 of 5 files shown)


src/main/scala/com/johnsnowlabs/nlp/annotators/common/TokenizedWithSentence.scala
```diff
@@ -24,7 +24,7 @@ object TokenizedWithSentence extends Annotated[TokenizedSentence] {
         token.begin >= sentence.start & token.end <= sentence.end
       ).map(token => IndexedToken(token.result, token.begin, token.end))
       sentenceTokens
-    }).filter(_.nonEmpty).zipWithIndex.map{case (indexedTokens, index) => TokenizedSentence(indexedTokens, index)}
```
saif-ellafi: Wait, i am not completely convinced, not sure if we want empty tokenized sentences here. Can't we remove the `zipWithIndex` and find out the original index from elsewhere?

PR #697, opened 2019-11-25
Chunk embeddings ref fix

jfernandrezj:
## Types of changes

## Checklist:

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,7 +5,6 @@ val scalaVer = "2.11.12"
 val scalaTestVersion = "3.0.0"
 
 val is_gpu = System.getProperty("is_gpu","false")
-
 /** Package attributes */
 if(is_gpu.equals("false")){
   name := "spark-nlp"
--- a/python/sparknlp/annotator.py
+++ b/python/sparknlp/annotator.py
@@ -494,6 +494,9 @@ class TextMatcherModel(AnnotatorModel):
             java_model=java_model
         )
 
+    def setMergeOverlapping(self, b):
+        return self._set(mergeOverlapping=b)
+
     @staticmethod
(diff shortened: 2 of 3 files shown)


src/main/scala/com/johnsnowlabs/nlp/embeddings/SentenceEmbeddings.scala
```diff
@@ -70,7 +70,11 @@ class SentenceEmbeddings(override val uid: String) extends AnnotatorModel[Senten
         begin = sentence.start,
         end = sentence.end,
         result = sentence.content,
-        metadata = Map.empty[String, String],
```
maziyarpanahi: We talked about this with @saif-ellafi since the content would be the entire sentence or worse the document, it is not a good idea to repeat this in metadata and add more memory/storage.
At the moment the SentenceEmbeddings has no annotator to use its metadata so we don’t know what should be the correct metadata. 
Maybe we can leave it empty for the time being?

maziyarpanahi: I know it’s coming from other PR, just thought we can discuss it here :)

saif-ellafi: Removing content from metadata requires a larger refactor. I suggest leaving this like it is now and let's work on that as a feature request for a breaking release.

maziyarpanahi: Instead of removing it, can we make the content empty? This will really increase the size of the DataFrame and there is no useful point to it.

PR #700, opened 2019-11-26
[WIP][BREAKING] All embeddings have mandatory ref, ref as a flexible trait

saif-ellafi:
## Description

Annotators that use Embeddings will now have a mandatory `embeddingsRef` so they only work with the embeddings they have been trained with.


saif-ellafi:
Deprecating this branch, continuing elsewhere

PR #741, opened 2020-01-13
Storage performance improvements

saif-ellafi:
## Description

Improvements in performance through Cache and reduced copy pasting


saif-ellafi:
Has issues

PR #754, opened 2020-01-24
Get the chunk metadata by its name rather than by its index

jfernandrezj:
## Description
ChunkEmbeddings was getting the chunk id by using head in the metadata (first key)
This PR changes it to get the explicit "chunk" key

## Motivation and Context

ChunkEmbeddings was failing on NER because the first index in an NER chunk is the actual entity

## How Has This Been Tested?
This is not a new feature
All tests passed

## Types of changes

## Checklist:

- [X] All new and existing tests passed.

### DIFF
--- a/src/main/scala/com/johnsnowlabs/nlp/embeddings/ChunkEmbeddings.scala
+++ b/src/main/scala/com/johnsnowlabs/nlp/embeddings/ChunkEmbeddings.scala
@@ -75,7 +75,7 @@ class ChunkEmbeddings (override val uid: String) extends AnnotatorModel[ChunkEmb
 
     val documentsWithChunks = annotations
       .filter(token => token.annotatorType == CHUNK)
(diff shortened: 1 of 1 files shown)


src/main/scala/com/johnsnowlabs/nlp/embeddings/ChunkEmbeddings.scala
```diff
@@ -75,7 +75,7 @@ class ChunkEmbeddings (override val uid: String) extends AnnotatorModel[ChunkEmb
 
     val documentsWithChunks = annotations
       .filter(token => token.annotatorType == CHUNK)
-      .groupBy(_.metadata.head._2.toInt)
+      .groupBy(_.metadata.getOrElse[String]("chunk", "0").toInt)
```
maziyarpanahi: @jfernandrezj Is there an issue in going by index?

PR #755, opened 2020-01-25
[WIP]Lazy recursive annotators

saif-ellafi:
## Description

1. RecursivePipeline -> Lets AnnotatorApproach that implement HasRecursiveFit to access other annotators in the pipeline (for example, tokenize a string with the same tokenizer that the user provided on the pipeline, for consistency)

2. RecursivePipelineModel -> Lets AnnotatorModel Access previous annotators in the pipeline through annotate() for example, to recalculate any results required on-demand. Examples are to creativity, but one example I was thinking is a LanguageRecognizer which can pick between multiple TokenizerModels to tokenize differently depending on the language.

3. LazyAnnotator -> All AnnotatorModel in a RecursivePipelineModel will not execute transform() or annotate() at all. It will only remain idle on the Pipeline for the RecursivePipeline accessor to annotate it manually on demand

And example use-case
```
  "Lazy Recursive Model" should "be ignored and used correctly once called" in {
    import com.johnsnowlabs.nlp.recursive._

    val lazyTokenizer = new Tokenizer().setInputCols("document").setOutputCol("token").setLazyAnnotator(true)

    val someModel = new SomeModel() {
      override def annotate(annotations: Seq[Annotation], recursivePipeline: Option[PipelineModel]): Seq[Annotation] = {

        val result = recursivePipeline.get.stages(1) match {
          case t: TokenizerModel => t.annotate(annotations)
          case _ => fail("Could not pattern match a TokenizerModel !")
        }

        // re-tokenize document annotations input
        result
      }

      override val inputAnnotatorTypes: Array[String] = Array(AnnotatorType.DOCUMENT)

    }.setInputCols("document").setOutputCol("baaar")
    val pipeline = new Pipeline().setStages(Array(document, lazyTokenizer, someModel))
    val output = pipeline.fit(data).recursive.transform(data)

    assert(output.schema.fields.length == 3)
    assert(output.schema.fields.map(_.name).sameElements(Array("text", "document", "baaar")))

    output.select("text", "document", "baaar").show()

    succeed
  }
```

Explanation of above example ^:

1. import of `recursive._` enables useful implicits to create `RecursivePipeline` and `RecursivePipelineModel`

2. `lazyTokenizer` has a param `LazyAnnotator` set to `true`

3. `someModel` is an annotator that implements `HasRecursiveTransform`

4. the pipeline model becomes recursive with `.recursive` statement

5. `someModeltakes` the recursive `lazyTokenizerand` manually tokenizes `DOCUMENT`

6. `lazyTokenizer` is completely ignored by the pipeline and `token` column will never exist


saif-ellafi:
Closed in favor of redesigned version:
https://github.com/JohnSnowLabs/spark-nlp/pull/756

PR #789, opened 2020-02-16
Chunk embeddings empty chunk fix

jfernandrezj:
Fix to get the chunk from the right place


maziyarpanahi:
Hi @jfernandrezj 
Could you please add the error, a code to reproduce the error and a unit test to make sure this PR resolves the error?

PR #790, opened 2020-02-17
Config prefix update to avoid -D option conflicts in Spark.

mptrepanier:
## Description
Prefixes the Spark NLP settings with `jsl.`.

## Motivation and Context
The current settings implementation of `sparknlp.settings. ...` leads to an exception being thrown by Spark's `validateSettings` when the Spark NLP configuration is provided as a `-D` option in either `spark.driver.extraJavaOptions` or `spark.executor.extraJavaOptions`, as shown here: 

https://github.com/apache/spark/blob/d8613571bc1847775dd5c1945757279234cb388c/core/src/main/scala/org/apache/spark/SparkConf.scala#L529

```
    getOption(executorOptsKey).foreach { javaOpts =>
      if (javaOpts.contains("-Dspark")) {
        val msg = s"$executorOptsKey is not allowed to set Spark options (was '$javaOpts'). " +
          "Set them directly on a SparkConf or in a properties file when using ./bin/spark-submit."
        throw new Exception(msg)
      }
      if (javaOpts.contains("-Xmx")) {
        val msg = s"$executorOptsKey is not allowed to specify max heap memory settings " +
          s"(was '$javaOpts'). Use spark.executor.memory instead."
        throw new Exception(msg)
      }
    }
```

Spark simply looks for any strings which contain `-Dspark` (which `-Dsparknlp` does) and errors out.

## How Has This Been Tested?
Against both a GCP Dataproc Cluster and a Zepplin notebook running on top of Dataproc. Example settings config shown below:

```
-Djsl.sparknlp.settings.license="LICENSE_KEY" -Djsl.sparknlp.settings.pretrained.credentials.access_key_id="ACCESS_KEY" -Djsl.sparknlp.settings.pretrained.credentials.secret_access_key="SECRET_ACCESS_KEY" -Djsl.sparknlp.settings.pretrained.s3_bucket=auxdata.johnsnowlabs.com -Djsl.sparknlp.settings.pretrained.s3_socket_timeout=3600000
```

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have read the [CONTRIBUTING](http://nlp.johnsnowlabs.com/contribute.html) page.

## Additional Info:

**IMPORTANT** The  `sparknlp.settings.license` references are not public. I believe these reside in the spark-nlp-license project and will need to be updated there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,9 +395,11 @@ ConfigLoader.setConfigPath("/somewhere/to/put/application.conf")
 And then we need to put in such application.conf the following content
 
 ```bash
-sparknlp {
-  settings {
-    cluster_tmp_dir = "somewhere in s3n:// path to some folder"
+jsl {
+  sparknlp {
+    settings {
+      cluster_tmp_dir = "somewhere in s3n:// path to some folder"
+    }
   }
 }
 ```
--- a/docs/en/quickstart.md
+++ b/docs/en/quickstart.md
@@ -162,11 +162,13 @@ ConfigLoader.setConfigPath("/somewhere/to/put/application.conf")
 And then we need to put in such application.conf the following content
 
 ```json
-sparknlp {
-    settings {
-        cluster_tmp_dir = "somewhere in s3n:// path to some folder"
+    jsl {
+        sparknlp {
+            settings {
+                cluster_tmp_dir = "somewhere in s3n:// path to some folder"
+            }
+        }
     }
-}
 ```
 
 For further alternatives and documentation check out our README page in [GitHub](https://github.com/JohnSnowLabs/spark-nlp).
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -1,29 +1,25 @@
-sparknlp {
-
-  settings {
-
-    overrideConfigPath = "./application.conf"
-    annotatorSerializationFormat = "object"
-    useBroadcastForFeatures = true
-
-    storage {
-      // Override with cluster location. Default is non set. Leads to fs.defaultFS scheme. e.g. hdfs:// in YARN.
-      // and hadoop.tmp.dir. Highly dependant on defaultFS. Override in non hadoop, e.g. S3 storage based clusters.
-      // cluster_tmp_dir = "s3://"
+jsl {
+  sparknlp {
+    settings {
+      overrideConfigPath = "./application.conf"
+      annotatorSerializationFormat = "object"
+      useBroadcastForFeatures = true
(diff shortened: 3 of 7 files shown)


maziyarpanahi:
@rohit-nlp I know you already fixed this in the Enterprise, do we really have to change everything in the public as well?  That validation code in Apache Spark has been there for the past 4 years, does this mean that none of our configs has ever worked? Or this is just to set them with -D or spark config?

rohit-nlp:
@maziyarpanahi our configs always worked because we used -DconfigFile=<location of application.conf file> and the settings was loaded using typeconfig and not spark. but when you pass some parameter of the configuration not through properties or application conf file but directly as java property such as -Dsparknlp.something then the spark code reading these params will throw an exception at this line.

https://github.com/apache/spark/blob/d8613571bc1847775dd5c1945757279234cb388c/core/src/main/scala/org/apache/spark/SparkConf.scala#L529


now as the code looks for "-Dspark" keyword -Dsparknlp will also get caugth in this.

maziyarpanahi:
@rohit-nlp So there is no need to rename anything here as users are supposed to pass those through DconfigFile or application.conf?

mptrepanier:
It comes down to whether or not you want to support the `-D` config options path. In the latest release, that was highlighted as one of the methods for setting SparkNLP config, but it currently errors out.

rohit-nlp:
I agree with @mptrepanier though there is a workaround with application conf we should support -D option directly for all properties too. Regarding --files not working in cluster mode is really surprising for me.. Though I did not get time to set up a yarn cluster and check the issue.  @maziyarpanahi we have already changed the property in Enterprise from sparknlp.settings.* to jsl.settings.* we can do the same change in public to be consistent.

maziyarpanahi:
That makes sense, @rohit-nlp do you approved this PR then?

mptrepanier:
> As well, the `--files` method provided by Spark does not distribute the `application.conf` file to the driver node, so the method:
> 
> ```
> spark-submit --files "$LOCAL_PATH/application.conf --conf spark.driver.extraJavaOptions="-Dconfig.file=file:application.conf" --conf spark.executor.extraJavaOptions="-Dconfig.file=file:application.conf" 
> ```
> 
> will fail in cluster mode as JSL's access key validation occurs in the driver.
> 
> (For local deployments `--master local[*]`, --files misleadingly will work as the driver and the executors are one.)

@rohit-nlp unable to replicate this.

PR #797, opened 2020-02-21
Fixed silent failure when version does not match

jfernandrezj:
Just a simple fix of raising the exception instead of returning it when tensorflow version doesnt match while creating the ner graph


maziyarpanahi:
@jfernandrezj we use TF 1.15 in 2.4.x and no longer require contrib. Why 1.12 and contrib? This looks like pre 2.4.x

PR #890, opened 2020-05-03
Annotator log file Fix for DBFS

rohit-nlp:
The Annotator log files where not getting created in DBFS due to IO Operation not supported error. This fix is to handle the error.
## Description

To make the fix the change was done on OutputHelper class where if FileSystem Scheme is of type "dbfs" it is treated as a normal disk file for creating directory and for writing and appending logs to the file.

The limitation of this approach is that the setOutputLogsPath in Annotators or sparknlp.settings.annotator.log_folder in application.conf can only point to a file system path and not the dbfs mount paths. File system paths like /home or /tmp of the driver. 

dbfs is a S3 like an object store which do not support appending to a file hence logs can not be created in dbfs paths. 

**Note:** 

1. If neither sparknlp.settings.annotator.log_folder is set nor setOutputLogsPath the log will go into /root/annotator_logs folder of the driver.
2. If the cluster in databricks is created with EBS volume type as "None" then the driver disk is not persisted on the restart and hence all the logs will be lost when the server is restarted. This is consistent behaviour as in with Driver Log in Spark UI.

## How Has This Been Tested?
 
The changes were tested in Databricks notebook for all the three approaches i.e 
1. Providing application.conf with sparknlp.settings.annotator.log_folder =  "/home/logs/"
2. Setting setOutputLogsPath for the annotator as "home/logNew"
3. Removing  sparknlp.settings.annotator.log_folder from application.conf which creates the log in the home folder i.e /root under annotator_logs folder

## Types of changes

## Checklist:

### DIFF
--- a/src/main/scala/com/johnsnowlabs/nlp/util/io/OutputHelper.scala
+++ b/src/main/scala/com/johnsnowlabs/nlp/util/io/OutputHelper.scala
@@ -5,23 +5,29 @@ import java.io.{File, FileWriter, PrintWriter}
 import com.johnsnowlabs.util.ConfigHelper
 import org.apache.hadoop.fs.{FileSystem, Path}
 
+
 object OutputHelper {
 
   lazy private val fs = FileSystem.get(ResourceHelper.spark.sparkContext.hadoopConfiguration)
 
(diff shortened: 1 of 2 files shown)


maziyarpanahi:
I'll merge and we test again once RC2 is released.

PR #957, opened 2020-06-30
Add support for Apache Spark 2.3.x

maziyarpanahi:
- [x] Allow compile/build/assembly with JVM option '-Dis_spark23=true`
- [x] Add Travis tests for Apache Spark 2.3 as well as Apache Spark 2.4
- [x] Adjust SparkNLP.start() and sparknlp.start() to accept `spark23` parameters to pull the appropriate artifacts for Apache Spark 2.3.x from Maven
- [x] Change ResourceDownloader and metadata.json o support multiple major versions


maziyarpanahi:
@saif-ellafi changing [2,4] to [2] in metadata.json doesn't seem to work in Apache Spark 2.3.x, have we changed the logic in ResourceDownloader to also check the second value in addition to the major version?

rohit-nlp:
changed the spark start functions a bit to change version only once instead of multiple times

rohit-nlp:
@maziyarpanahi changing to 2,3 should work too for 2,3,* and 2,4,* at least for version compatibility test..
assert(Version.isCompatible(Version(2,3,4), Version(2,3))) ==> True
assert(Version.isCompatible(Version(2,4,4), Version(2,3))) ==> True
assert(Version.isCompatible(Version(2,3,4), Version(2))) ==> True
assert(Version.isCompatible(Version(2,4,4), Version(2))) ==> True

maziyarpanahi:
> @maziyarpanahi changing to 2,3 should work too for 2,3,* and 2,4,* at least for version compatibility test..
> assert(Version.isCompatible(Version(2,3,4), Version(2,3))) ==> True
> assert(Version.isCompatible(Version(2,4,4), Version(2,3))) ==> True
> assert(Version.isCompatible(Version(2,3,4), Version(2))) ==> True
> assert(Version.isCompatible(Version(2,4,4), Version(2))) ==> True

That's very interesting! So we can interpret the Version as the minimum required/compatible Apache Spark major release? Like saying (2,3) means it works with 2,3 and 2,4. Out of curiosity, what if the user's Apache Spark is (2,5) or (3,0)?

Another question, when I do change the Version in metadata.json for a model to `"sparkVersion":{"parts":[2]}` it doesn't download it in Apache Spark 2.3.x, is that normal?

rohit-nlp:
> > @maziyarpanahi changing to 2,3 should work too for 2,3,* and 2,4,* at least for version compatibility test..
> > assert(Version.isCompatible(Version(2,3,4), Version(2,3))) ==> True
> > assert(Version.isCompatible(Version(2,4,4), Version(2,3))) ==> True
> > assert(Version.isCompatible(Version(2,3,4), Version(2))) ==> True
> > assert(Version.isCompatible(Version(2,4,4), Version(2))) ==> True
> 
> That's very interesting! So we can interpret the Version as the minimum required/compatible Apache Spark major release? Like saying (2,3) means it works with 2,3 and 2,4. Out of curiosity, what if the user's Apache Spark is (2,5) or (3,0)?
> 
> Another question, when I do change the Version in metadata.json for a model to `"sparkVersion":{"parts":[2]}` it doesn't download it in Apache Spark 2.3.x, is that normal?

[2,3] will work for [2,4,*] [2,5,*] but not for 3

rohit-nlp:
Just changing meta data wont work need to change the zip file names too

rohit-nlp:
Ok so if we want to support pre-trained models currently tagged for 2.4 to work also for 2.3 there are two approaches. Both will require metadata changes

1. Approach 1: in metadata.json update spark version to [2,3] so that it will work for 2.4 as well and add a new tag as file name with the name of the zip file. Now the resource downloader will use the filename in the metadata json to look up for file if compatibility is found instead of constructing the path based on name, lib version, spark version, timestamp as done now. Changes required: Metadata.json, ResourceDownloader code and one-time script to update the metadata of all the files.

2. Approach 2: in metadata.json update spark version to [2,3] so that it will work for 2.4  and rename all the files to be _2_3 instead of _2_4. This can be done using a simple python script. 

@maziyarpanahi @saif-ellafi  vote?

maziyarpanahi:
> Ok so if we want to support pre-trained models currently tagged for 2.4 to work also for 2.3 there are two approaches. Both will require metadata changes
> 
> 1. Approach 1: in metadata.json update spark version to [2,3] so that it will work for 2.4 as well and add a new tag as file name with the name of the zip file. Now the resource downloader will use the filename in the metadata json to look up for file if compatibility is found instead of constructing the path based on name, lib version, spark version, timestamp as done now. Changes required: Metadata.json, ResourceDownloader code and one-time script to update the metadata of all the files.
> 2. Approach 2: in metadata.json update spark version to [2,3] so that it will work for 2.4  and rename all the files to be _2_3 instead of _2_4. This can be done using a simple python script.
> 
> @maziyarpanahi @saif-ellafi vote?

@rohit-nlp @saif-ellafi Which option leaves us with enough flexibility to anticipate the upcoming Scala 2.12 and Apache Spark 3.0 (knowing the models for these two won't work on 2.3 and 2.4)?

rohit-nlp:
> > Ok so if we want to support pre-trained models currently tagged for 2.4 to work also for 2.3 there are two approaches. Both will require metadata changes
> > 
> > 1. Approach 1: in metadata.json update spark version to [2,3] so that it will work for 2.4 as well and add a new tag as file name with the name of the zip file. Now the resource downloader will use the filename in the metadata json to look up for file if compatibility is found instead of constructing the path based on name, lib version, spark version, timestamp as done now. Changes required: Metadata.json, ResourceDownloader code and one-time script to update the metadata of all the files.
> > 2. Approach 2: in metadata.json update spark version to [2,3] so that it will work for 2.4  and rename all the files to be _2_3 instead of _2_4. This can be done using a simple python script.
> > 
> > @maziyarpanahi @saif-ellafi vote?
> 
> @rohit-nlp @saif-ellafi Which option leaves us with enough flexibility to anticipate the upcoming Scala 2.12 and Apache Spark 3.0 (knowing the models for these two won't work on 2.3 and 2.4)?

a model tagged as 2,3 will pass version compatibility with 2,4,* but will not (and should not) pass compatibility check with 3.*

for scala its a good point currently we check the compatibility of spark version, spark-nlp lib version but not scala version we should put it in metadata too if for spark 2,4 we are planning to support both scala 2.11 and scala 2,12

maziyarpanahi:
> > > Ok so if we want to support pre-trained models currently tagged for 2.4 to work also for 2.3 there are two approaches. Both will require metadata changes
> > > 
> > > 1. Approach 1: in metadata.json update spark version to [2,3] so that it will work for 2.4 as well and add a new tag as file name with the name of the zip file. Now the resource downloader will use the filename in the metadata json to look up for file if compatibility is found instead of constructing the path based on name, lib version, spark version, timestamp as done now. Changes required: Metadata.json, ResourceDownloader code and one-time script to update the metadata of all the files.
> > > 2. Approach 2: in metadata.json update spark version to [2,3] so that it will work for 2.4  and rename all the files to be _2_3 instead of _2_4. This can be done using a simple python script.
> > > 
> > > @maziyarpanahi @saif-ellafi vote?
> > 
> > 
> > @rohit-nlp @saif-ellafi Which option leaves us with enough flexibility to anticipate the upcoming Scala 2.12 and Apache Spark 3.0 (knowing the models for these two won't work on 2.3 and 2.4)?
> 
> a model tagged as 2,3 will pass version compatibility with 2,4,* but will not (and should not) pass compatibility check with 3.*
> 
> for scala its a good point currently we check the compatibility of spark version, spark-nlp lib version but not scala version we should put it in metadata too if for spark 2,4 we are planning to support both scala 2.11 and scala 2,12 or we retrain models for spark 2_4 with scala 2_12 and upload them as model_name_212 for spark version 2_4 as spark 2.4 is the only version which supports both scala 211 and 212

We can support Scala 2.12 only for Apache Spark 3.x which is the only possibility and that way we don't have to incorporate the Scala version in our logic. The 2,3 and 2,4 are in Scala 2.11 and they are cross-compatible so we are good there, and Spark 3,x will not be used in 2,x anyway so we are good there.

I think option number one which doesn't require renaming (re-uploading) is the best option. (anything that doesn't require renaming/reuploading 160 models, the code and metadata changes are easy)

rohit-nlp:
@maziyarpanahi ok let me do this change in this same branch then and I will commit it soon (1-2 days max) will that be ok with our next minor release plan?

maziyarpanahi:
> @maziyarpanahi ok let me do this change in this same branch then and I will commit it soon (1-2 days max) will that be ok with our next minor release plan?

Yes, that would be perfect! I want to release it so we can do more tests before officially saying we fully support 2.3 as well as 2.4.

PR #1085, opened 2020-10-05
[WIP] Batch Annotators

saif-ellafi:
## Description

**NOT ready for MERGE**
This PR proposes Batch Annotators interface, it allows for annotators to deal with their own algorithms in batches of Rows, rather row per row. Some annotators like `NerDL` benefit from this since they make external API calls to Tensorflow. Larger batches here show up to ~3 times faster performance speed than without batching multiple rows at once.

## Types of changes

## Checklist:


vkocaman (changes requested):
@saif-ellafi I tested this jar and looks great.. but we also see some issue with validation step after each epoch.. and it is still doing the valuation row by row.. we start the ner training, it grabs all the cpus on the machine and then while evaluting for val, it drops to single core.. can you do the same fix on that part as well ?

src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorflowNer.scala

Epoch 1/50 started, lr: 0.003, dataset size: 49025
Epoch 1/50 - 179.65s - loss: 530.2455 - batches: 51
Quality on validation dataset (20.0%), validation examples = 9805
time to finish evaluation: 1117.67s

maziyarpanahi:
> @saif-ellafi I tested this jar and looks great.. but we also see some issue with validation step after each epoch.. and it is still doing the valuation row by row.. we start the ner training, it grabs all the cpus on the machine and then while evaluting for val, it drops to single core.. can you do the same fix on that part as well ?
> 
> src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorflowNer.scala
> 
> Epoch 1/50 started, lr: 0.003, dataset size: 49025
> Epoch 1/50 - 179.65s - loss: 530.2455 - batches: 51
> Quality on validation dataset (20.0%), validation examples = 9805
> time to finish evaluation: 1117.67s

There is confusion between a single row and a single row. 

- A row per annotation is not done on a single core, it is still multi-core/multi-thread
- The work here is done for Transform not Train, so whatever happens during the training .fit() is not included in the BatchAnnotation
- What you see is what @albertoandreottiATgmail has done in the new NerDL in 2.6.3, I've never seen a single core issue on the evaluation of NerDL during the training. Is `enableMemoryOptimizer` enabled by any chance?

saif-ellafi:
Hello,
So this branch is an experiment, what it does is basically:

1. Hey Spark Executor, collect all the rows you have (Up to a batch size limit)
2. Give them to the NerDLModel (only prediction is applied for now) all together.
3. NerDLModel would treat them as batched sentences
4. Give it back to the planner (a MapPartitions planner) and split them back again into rows

The outcome of this, is that the information sent to Tensorflow is batched in larger groups.

The results were significantly good, in NerDLModel, however, in other Tensorflow annotators like BERT, the overhead was heavier than the performance gained for batching, and for this reason is that the PR is a bit on pause. It was discarded that all Tensorflow annotators would benefit from batching, more as row-per-row.

You can close the PR as for now there is no apparent alternative until more conclusive results show up, or of course if anyone else has any ideas.

vkocaman:
@maziyarpanahi yes , eval taking time issue only happens in 2.6.3.. it is fine at 2.6.2

2.6.2 1k+ seconds
2.6.2 60 seconds

But, we still see a single core grabbed while evaluating.. 

@saif-ellafi we tested this branch  in a real project and it looks really good..  thx again

maziyarpanahi:
> @maziyarpanahi yes , eval taking time issue only happens in 2.6.3.. it is fine at 2.6.2
> 
> 2.6.2 1k+ seconds
> 2.6.2 60 seconds
> 
> But, we still see a single core grabbed while evaluating..
> 
> @saif-ellafi we tested this branch in a real project and it looks really good.. thx again

Great, so I changed the label on the issue as a bug, it is probably not respecting that the memory optimization is off for evaluation.

vkocaman:
@saif-ellafi @maziyarpanahi I recently tested this batch annotation on 100 large docs but ended up 142 parsed docs in return.. so it makes me think that while batching, we are somehow replicating the records ? I know it sounds weird and will debug if I can.

saif-ellafi:
> @saif-ellafi @maziyarpanahi I recently tested this batch annotation on 100 large docs but ended up 142 parsed docs in return.. so it makes me think that while batching, we are somehow replicating the records ? I know it sounds weird and will debug if I can.

Thanks for the testing, that explains why could it be that it appears to be slower than it should. Will take a look when possible at the batch-merging logic.

maziyarpanahi:
> > @saif-ellafi @maziyarpanahi I recently tested this batch annotation on 100 large docs but ended up 142 parsed docs in return.. so it makes me think that while batching, we are somehow replicating the records ? I know it sounds weird and will debug if I can.
> 
> Thanks for the testing, that explains why could it be that it appears to be slower than it should. Will take a look when possible at the batch-merging logic.

Many thanks @saif-ellafi, we also fixed a few stuff in padding inside TensorflowBert that might have made the BERT equal or slower when ti comes to BatchAnnotation.

vkocaman:
> > @saif-ellafi @maziyarpanahi I recently tested this batch annotation on 100 large docs but ended up 142 parsed docs in return.. so it makes me think that while batching, we are somehow replicating the records ? I know it sounds weird and will debug if I can.
> 
> Thanks for the testing, that explains why could it be that it appears to be slower than it should. Will take a look when possible at the batch-merging logic.

@saif-ellafi Thanks.. I also monitored the cpu loads and noticed that most of the time only one cpu is loaded 100%, and then all the cores max out a few seconds and then a few minutes work on a single core.. just wanted to share

saif-ellafi:
> > > @saif-ellafi @maziyarpanahi I recently tested this batch annotation on 100 large docs but ended up 142 parsed docs in return.. so it makes me think that while batching, we are somehow replicating the records ? I know it sounds weird and will debug if I can.
> > 
> > 
> > Thanks for the testing, that explains why could it be that it appears to be slower than it should. Will take a look when possible at the batch-merging logic.
> 
> @saif-ellafi Thanks.. I also monitored the cpu loads and noticed that most of the time only one cpu is loaded 100%, and then all the cores max out a few seconds and then a few minutes work on a single core.. just wanted to share

For training... that is reasonable.. but for prediction, it should not. Regardless of batch or non-batch. If your dataframe is multiple partitions and you are running predictions with `--master local[16]` or something alike, it should spread the cpu usage to different partitions.

vkocaman:
> > > > @saif-ellafi @maziyarpanahi I recently tested this batch annotation on 100 large docs but ended up 142 parsed docs in return.. so it makes me think that while batching, we are somehow replicating the records ? I know it sounds weird and will debug if I can.
> > > 
> > > 
> > > Thanks for the testing, that explains why could it be that it appears to be slower than it should. Will take a look when possible at the batch-merging logic.
> > 
> > 
> > @saif-ellafi Thanks.. I also monitored the cpu loads and noticed that most of the time only one cpu is loaded 100%, and then all the cores max out a few seconds and then a few minutes work on a single core.. just wanted to share
> 
> For training... that is reasonable.. but for prediction, it should not. Regardless of batch or non-batch. If your dataframe is multiple partitions and you are running predictions with `--master local[16]` or something alike, it should spread the cpu usage to different partitions.

I am talking about prediction.. .I just tried getting predictions after reading from a parquet and still same..

PR #1184, opened 2020-11-21
Introducing a new DocumentNormalizer annotator

maziyarpanahi:
This PR introduces a new annotator that normalizes raw text from the tagged text, e.g. scraped web pages or XML documents, from document type (output of DocumentAssembler) to document type.
  * Removes all dirty characters from text following one or more input regex patterns.
  * Can apply non-wanted character removal which a specific policy.
  * Can apply lower case normalization.


albertoandreottiATgmail:
@maziyarpanahi can we have somewhere the ability to restrict the encoding to UTF-8 only?

maziyarpanahi:
That's a great idea, what about having "encode" and let users decide the encoder?

albertoandreottiATgmail:
That may work. I understand Java Strings use UTF-16... so my next question would be what about all the other stuff we put in the pipeline after this annotator?
Probably when training new stuff you will just have everything learn from utf-8, but there's a chance of some incompatibility with pretrained stuff, not really sure about this myself.

maziyarpanahi:
That's true, all the trainable annotators get their training data from outside of the Spark NLP pipeline. Even the CoNLL() class won't allow any settings for any params. But at least it would be good to have the option to set `encoding` for those who wish to change it for some reason if they need an English model but might have other characters that they may not be interested in. 
I'll ask around to see if this is interesting since this annotator can be a good place for this kind of stuff.

maziyarpanahi:
@albertoandreottiATgmail Could you please review the new encoding param? You have more experience in Java than me. Also, should we leave the default be no encoding rather than `UTF-8`?

PR #2186, opened 2021-01-23
TF2 support

albertoandreottiATgmail:
## Annotator models that compile and run successfully
The result of the model is not important in this section, we care about the compatibility of the model exported in TF v1 and being able to load in TF v2:

- [x] Albert
- [x] Bert
- [x] BertSentence
- [x] Elmo
- [ ] USE
- [ ] Xlnet
- [ ] MarianTransformer

## The models which were tested based on their results:
We care about the result of the model, whether or not the embeddings are the same, translations are the same, etc.

### Evaluation failed 
- MarianTransformer is not outputting desired translation in the unit test. There must be an issue with `offset` and the new NdArray buffers which are different than Java buffers with `flip`


maziyarpanahi:
Merging this to Spark 3.0 support to make the maintaining easier.

PR #2464, opened 2021-03-16
[skip ci] 2021-03-16-sentence_detector_dl_healthcare_en

jsl-models:


vkocaman (changes requested):
We need to use the respective hc notebook and relevant examples.. you can just copy paste form this nb and use as an example

https://github.com/JohnSnowLabs/spark-nlp-workshop/blob/master/tutorials/Certification_Trainings/Healthcare/20.SentenceDetectorDL_Healthcare.ipynb

PR #2619, opened 2021-03-29
2021-03-29-pos_fast_en

jsl-models:


vkocaman:
@C-K-Loan  why is this licensed ?

C-K-Loan:
@vkocaman  I found this model in the S3 bucket for licensed healthcare models. So I assume this was intended for licensed users, but maybe it was also just a prototype for something and left undocumented

vkocaman:
> @vkocaman I found this model in the S3 bucket for licensed healthcare models. So I assume this was intended for licensed users, but maybe it was also just a prototype for something and left undocumented.
> If nobody knows what this model is used for or what's up with it we can just close this pull

yep, lets close this up.. it can be pushed to open source bucket though ?

maziyarpanahi:
Thank you all, this model is very old and deprecated. We used a very small dataset to train it.

PR #2680, opened 2021-04-01
2021-04-01-ner_bionlp_noncontrib_en

jsl-models:


vkocaman:
We don't need non-contrib versions anymore..

PR #2708, opened 2021-04-01
2021-04-01-visual_document_classifier_tobacco_3482_en

jsl-models:


vkocaman:
@kolia1985 I suggest you add the full piepline in the example code.. starting from taking the image to feeding to classifier. ImageToHocr + VisualDocClassifer

maziyarpanahi:
Should’d the edition be Spark OCR instead of Spark NLP? We don’t have such annotator in the public library.

vkocaman:
@maziyarpanahi this should be closed I guess.. I asked this @kolia1985 before and he said that it was a test PR to test something

PR #5666, opened 2021-06-06
2021-06-06-ner_deid_subentity_glove_en

jsl-models:


vkocaman:
@murat-gunay word emb and MedicalNer both need to use sentence or document in the input col

murat-gunay:
> @murat-gunay word emb and MedicalNer both need to use sentence or document in the input col

@vkocaman hi, input col was corrected as you mentioned

PR #5677, opened 2021-06-08
Feature/allow stdout sterr pipes

danilojsl:
Allow access to stdout/stderr for Python notebooks

## Description

Modifies sparknlp.start() to include parameters that enable outputting stdout/stderr in real-time for child processes. The change required to lunch a custom gateway when starting a spark session.
This option is available for Spark 3.0.x +

## Motivation and Context

Currently, every log and stdout/stderr inside Java/Scala will be only available in the parent process via py4j. For instance, during the training of NerDLApproach, the println only shows up in the console log.
However, in every single Python notebook (Jupyter, Zepplin, Databricks, etc.) this won't be the output of the notebook rather than the runtime logs.
Thus, we need a way to have access to stdout/stderr via JVM with a minimum amount of code changes.

## How Has This Been Tested?

Local tests on Jupyter notebook and the following Google Colab Notebooks:
[SparkNLP - Training NER Real Time Output](https://colab.research.google.com/drive/1neOcUYYWZm7v0HTvleXj_Y42_7L8o2LY?usp=sharing) (New Behaviour)
[SparkNLP - Training NER](https://colab.research.google.com/drive/1LrnPCB8LQqZKb_3Wya66CVgYyB56dWuq?usp=sharing) (Current Behaviour)

## Types of changes

## Checklist:


maziyarpanahi:
Hi @danilojsl 
This is really good! it is going to be very useful for Jupyter and Colab notebooks.

One question since I know it will come up right after we release this, what should be done in environments when the SparkSession is created automatically and all users can do is to add some configurations. Like:
- Databricks
- Zeppelin on EMR, Cloudera, and Hortonworks

Is it possible to have some instructions or it's not possible at all?

danilojsl:
Hi @maziyarpanahi, I'm afraid it won't be possible in the aforementioned big data platforms since we require to define a custom gateway when creating a Spark Context with PySpark. 
To the best of my knowledge, that configuration is not accessible for users of those platforms.  Having said that, two options that we have:

1. Perhaps users of those platforms can open a ticket or ask their admins to see if they can do something as we do on our `__init__.py` file when starting a Spark Session.
2. Ask Apache Spark for a feature in PySpark. It could be a function to set gateway configuration for a running Spark Context. So, users of those data platforms can set this value when accessing the created Spark Context.

maziyarpanahi:
> Hi @maziyarpanahi, I'm afraid it won't be possible in the aforementioned big data platforms since we require to define a custom gateway when creating a Spark Context with PySpark.
> To the best of my knowledge, that configuration is not accessible for users of those platforms. Having said that, two options that we have:
> 
> 1. Perhaps users of those platforms can open a ticket or ask their admins to see if they can do something as we do on our `__init__.py` file when starting a Spark Session.
> 2. Ask Apache Spark for a feature in PySpark. It could be a function to set gateway configuration for a running Spark Context. So, users of those data platforms can set this value when accessing the created Spark Context.

Thanks @danilojsl 
I think the second option should be the case. If there is no risk or any overhead for defining a custom gateway automatically in Spark Context then it should be added by the Apache Spark team. I'll look to see if there is any ticket already for this. (they might have added something to support this in Spark 3.x)

maziyarpanahi (approved):

PR #5739, opened 2021-06-28
Change default value for explodeSentences in CoNLL class to false

maziyarpanahi:
In previous releases, this parameter was set to `true` to explode sentences for better distribution/parallelism. However, thanks to batchAnnotate being implemented in the most popular Word Embeddings annotator, not exploding the sentences will massively utilize GPU/CPU which speeds up the transformation/collect needed inside `NerDLApproach`.


maziyarpanahi:
This may not be safe for CoNLL files where there are no boundaries for DOCUMENT. In which it will put everything in 1 row! Let's leave this to the users to decide whether they want this param to be True or False.

PR #5771, opened 2021-07-04
Add includeAllConfidenceScores param to NerDLModel and NerDLApproach

maziyarpanahi:
This param allows outputting only the confidence score for the predicted label instead of all labels. In some cases with lots of NER tags, this param helps to improve the speed. (by default it's set to false)

### DIFF
--- a/python/sparknlp/annotator.py
+++ b/python/sparknlp/annotator.py
@@ -1553,6 +1553,10 @@ class NerDLApproach(AnnotatorApproach, NerApproach):
                               "whether to include confidence scores in annotation metadata",
                               TypeConverters.toBoolean)
 
+    includeAllConfidenceScores = Param(Params._dummy(), "includeAllConfidenceScores",
+                                       "whether to include all confidence scores in annotation metadata or just the score of the predicted tag",
+                                       TypeConverters.toBoolean)
+
     enableOutputLogs = Param(Params._dummy(), "enableOutputLogs",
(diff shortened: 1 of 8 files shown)


danielle-io:
Hello! I came across this PR when investigating an issue I'm having loading the pretrained pipeline "recognize_entities_dl." I've used this pipeline in the past without issue, but started getting an error when I tried to download the pipeline today - "Param includeAllConfidenceScores does not exist." This doesn't happen when I switch the pipeline to "explain_document_ml." I thought I'd comment here in case this isn't something I should open an issue for.

The line of code that triggers the error:
pipeline = PretrainedPipeline('explain_document_ml', 'en')

The pipeline is still up at https://nlp.johnsnowlabs.com/docs/en/pipelines#recognizeentitiesdl so I assume it's still supported, but if I'm wrong, my apologies!

maziyarpanahi:
Hi @danielle-io 

Many thanks, I will look into it. However, make sure both PyPI and Maven/JAR in your environment are 3.1.2, this usually happens when one is ahead of the other one.

danielle-io:
@maziyarpanahi thanks for the quick response! They’re both on 3.1.2, I double checked now just in case. The second pipeline I mentioned works without any issues (explain_doc), it’s just the entity pretrained pipeline that triggered the confidence scores error.

danielle-io:
@maziyarpanahi youre right I saw 2.12 and 3.0.1 and my brain mixed them up 🙈 thank you!

PR #5824, opened 2021-07-15
2021-07-15-explain_clinical_doc_ade_en

jsl-models:


vkocaman:
@HashamUlHaq  we could have a better example to cover all the pipeline components.. and `is gone` shouldn't be the part of the chunk

PR #5853, opened 2021-07-21
2021-07-21-lener_bert_large_pt

jsl-models:


vkocaman:
@dcecchini any NER model uploaded to healthcare should be trained by MedicalNER so that people wouldn't use that with open source NerDL.. in order not to retrain, you can ask @vankov to help you convert this into a licensed one for once. 

I merged the previous one without noticing this.. so you should do the same for that model as well.

dcecchini:
I contacted @vankov  and will wait for instructions on how to change the ner model for the licensed version.

dcecchini:
@vkocaman I re-uploaded the models and the PRs will be different. Please consider these ones: [lener_bert_base](https://github.com/JohnSnowLabs/spark-nlp/pull/5882) and [lener_bert_large](https://github.com/JohnSnowLabs/spark-nlp/pull/5883)

PR #5971, opened 2021-08-23
Update LICENSE

masguit42:
## Types of changes

## Checklist:

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright {yyyy} {name of copyright owner}
+   Copyright 2021, John Snow Labs.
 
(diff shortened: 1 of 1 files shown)


maziyarpanahi:
Thanks @masguit42 for the contribution. That part of the license is a template for those who wish to use the licensed as a header in their own work: “APPENDIX: How to apply the Apache License to your work.”

It should stay the same for instance: https://github.com/apache/spark/blob/master/LICENSE

maziyarpanahi:
On a second thought maybe that could be what it is for others to use as example in their projects. I will ask.

maziyarpanahi:
OK it's definitely a template for others: https://github.com/scala/scala/blob/2.13.x/LICENSE

However, we might ourselves use that template in our own license when we include them as a header in the source code. I can discuss this to see if we have to.

@masguit42 many thanks again to bring this to our attention 👍🏼

PR #6044, opened 2021-08-31
Preserve existing metadata in BertSentenceEmbeddings

maziyarpanahi:
This PR extends the Sentence case class with metadata to be used when necessary to preserve metadata generated by previous annotators.

### DIFF
--- a/src/main/scala/com/johnsnowlabs/nlp/annotators/common/SentenceSplit.scala
+++ b/src/main/scala/com/johnsnowlabs/nlp/annotators/common/SentenceSplit.scala
@@ -18,15 +18,17 @@ package com.johnsnowlabs.nlp.annotators.common
 
 import com.johnsnowlabs.nlp.{Annotation, AnnotatorType}
 
+import scala.collection.Map
+
 /**
-  * structure representing a sentence and its boundaries
-  */
-case class Sentence(content: String, start: Int, end: Int, index: Int)
+ * structure representing a sentence and its boundaries
+ */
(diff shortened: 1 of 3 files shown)


src/main/scala/com/johnsnowlabs/nlp/annotators/common/SentenceSplit.scala
```diff
@@ -18,15 +18,17 @@ package com.johnsnowlabs.nlp.annotators.common
...
 
+import scala.collection.Map
+
 /**
-  * structure representing a sentence and its boundaries
-  */
-case class Sentence(content: String, start: Int, end: Int, index: Int)
+ * structure representing a sentence and its boundaries
+ */
+case class Sentence(content: String, start: Int, end: Int, index: Int, metadata: Map[String, String] = null)
```
xusliebana, line 26: Maybe is more usefull use Option[Map[String,String]]  it is a safer way to avoid nullPointerExceptions.
case class Sentence(content: String, start: Int, end: Int, index: Int, metadata: Map[String, String] = None)

maziyarpanahi: Is this better? (both the use of Option in SentenceSplit and TensorflowBert)

xusliebana: Yes I think that is more scala idiomatic.

xusliebana (approved):

PR #6084, opened 2021-09-08
2021-09-08-classifierdl_bert_sentiment_fr

jsl-models:


murat-gunay:
`SentimentDLApproach()` gave not as good benchmark as `ClassifierDLApproach()`. So I trained the model using `ClasifierDLApproach()`

PR #6099, opened 2021-09-13
2021-09-13-bert_token_classifier_ner_deid_en

jsl-models:


maziyarpanahi:
@murat-gunay @vkocaman Is this supposed to be a public model?

murat-gunay:
@maziyarpanahi  i wrongly uploaded it as public but it should be licensed. i am thinking whether i should close this PR and do it again or I can fix it only correcting md file?

maziyarpanahi:
@murat-gunay it's already uploaded in the wrong place with all kinds of wrong metadata on S3, it's not just the markdown. I think re-uploading is the safest and easiest path.

PR #6156, opened 2021-09-24
2021-09-24-bert_token_classifier_ner_jsl_slim_en

jsl-models:


murat-gunay:
hi @vkocaman  and @HashamUlHaq 
- this is bert4token version of `ner_jsl_slim` model.
- I already shared the metrics comparison of this model with @vkocaman 
- The prediction of the real texts is quite good.

PR #6181, opened 2021-09-28
Update ResourceDownloader.showPublicPipelines interface and README.md

DevinTDHa:
## Description

- updated `ResourceDownloader.showPublicPipelines` to be more in line with the `ResourceDownloader.showPublicModels` function
- providing no version will use the current version to check for compatibility
- updated README.md to show examples on how to use those functions

## How Has This Been Tested?

Added tests for `ResourceDownloader.showPublicPipelines`.

## Types of changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -1033,6 +1033,49 @@ annotation.select("entities.result").show(false)
 */
 ```
 
+#### Showing Available Pipelines
+
+There are functions in Spark NLP that will list all of the available Pipelines
+of a particular language for you:
+
+```scala
+import com.johnsnowlabs.nlp.pretrained.ResourceDownloader
+
+ResourceDownloader.showPublicPipelines(lang="en")
+/*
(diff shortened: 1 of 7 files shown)


maziyarpanahi:
@DevinTDHa 
This is great, can we have also Python examples (the import and the function so they can copy/paste)? Also, one example of how to use an annotator's name with version to see all the available models. (same for pipelines)

DevinTDHa:
> @DevinTDHa This is great, can we have also Python examples (the import and the function so they can copy/paste)? Also, one example of how to use an annotator's name with version to see all the available models. (same for pipelines)

I have added the examples and fixed the python README.md so it only has python code now.

Additionally, commit 3c82fde fixes some issues, regarding the python interface. It would work in the python test but the the output wouldn't be properly forwarded to notebooks.

maziyarpanahi (approved):

PR #6369, opened 2021-11-02
2021-11-02-snomed_sentence_resolver_es

jsl-models:


vkocaman:
@josejuanmartinez it would be better to follow the naming conventions for resolvers..  for example, we call english version as `sbertresolve_snomed_conditions` . So you may want to name this as `robertaresolve_snomed`, and if it is just using conditions `robertaresolve_snomed_conditions`.

josejuanmartinez:
I close this PR and create a new one with the proper file name.

PR #6370, opened 2021-11-03
2021-11-03-bert_token_classifier_ner_bionlp_en

jsl-models:


vkocaman:
@murat-gunay can you compare these metrics with https://nlp.johnsnowlabs.com/2021/03/31/ner_bionlp_en.html
it looks like a weak model..

murat-gunay:
Hi @vkocaman 
I trained several versions playing with hyperparameters under the limitations of GPUzilla. Yes, It is not a powerful model but I compared to the `ner_bionlp` and it is not also a powerful model which has 0.63 macro. There is no big difference when I get predictions. Works well.

murat-gunay:
@vkocaman i trained it using different hyperparameter on colab pro. the results are a little bit better. So I am closing this PR and re-upload it again.

PR #6385, opened 2021-11-03
2021-11-03-bert_token_classifier_ner_bionlp_en

jsl-models:


murat-gunay:
@vkocaman hi
this is a little bit bettter than previous one trained on other than GPUzilla. the improvement is %3

PR #6389, opened 2021-11-03
2021-11-03-ner_profiling_clinical_en

jsl-models:


vkocaman:
@muhammetsnts  the example should also be the chunk based.. no need to write all, but the outputs from a few ner models would be enough.. I assume you updated the notebook as well.

PR #6446, opened 2021-11-10
Multicolumns approach

xusliebana:
Add an interface to use multiple input annotations columns

## Motivation and Context

We need to have an interface to allow to use of multiple input annotations in the healthcare repo.

## How Has This Been Tested?
I created a unit test that show that is working that approach

## Types of changes

- [X ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ X] My code follows the code style of this project.


maziyarpanahi:
@xusliebana the references to `CHUNK` in test should change to `DOCUMENT`.

xusliebana:
@maziyarpanahi Done

maziyarpanahi (approved):

maziyarpanahi:
Thanks @xusliebana 
This will be in the next release of `spark-nlp==3.3.3`

src/main/scala/com/johnsnowlabs/nlp/HasMultipleInputAnnotationCols.scala
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.johnsnowlabs.nlp
+
+/**
+ * Trait  used to create annotators with input columns of variable length.
+ * */
+trait HasMultipleInputAnnotationCols extends HasInputAnnotationCols {
```
albertoandreottiATgmail, line 22: @maziyarpanahi @xusliebana , don't you guys think that the name for this trait could be `HasVariableCountInputAnnotationCols`

HasInputAnnotationCols already supports multiple input columns, the main change I see here is that the number of columns will no longer be defined in a 'declarative' fashion inside the library, but in a more dynamic way by the user, during annotator setup.

maziyarpanahi: @albertoandreottiATgmail Yes, that totally makes sense! I'll leave it to you and @xusliebana to rename this and the unit tests. (@xusliebana you can directly fix the changes inside release/333-release-candidate branch)

@albertoandreottiATgmail Could you please also review removing `final` from `final def setInputCols()`? I am not sure why we used `final` so I want to be sure removing it won't be risky in some scenarios.

PR #6502, opened 2021-11-22
NMH-34: Add fields to filtering functionality [skip test]

KshitizGIT:
Jira Ticket [NMH-34](https://johnsnowlabs.atlassian.net/browse/NMH-34)

### DIFF
--- a/docs/_plugins/search_index.rb
+++ b/docs/_plugins/search_index.rb
@@ -74,6 +74,11 @@ class Extractor
     m ? m[1] : nil
   end
 
+  def doc_type
+    m = /\|Type:\|(.*?)\|/.match(@content)
+    m ? (m[1] == 'pipeline' ? m[1]: 'model') : nil
+  end
+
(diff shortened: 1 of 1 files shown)


docs/_plugins/search_index.rb
```diff
@@ -74,6 +74,11 @@ def download_link
     m ? m[1] : nil
   end
 
+  def model_type
```
pabla, line 77: We need only two possible results here: `model` or `pipeline`.

pabla (changes requested):

docs/_plugins/search_index.rb
KshitizGIT: what if the docs, don't have the Type ?

KshitizGIT: Updated

pabla (approved):

pabla:
All changes from this PR were applied in #6610

PR #6523, opened 2021-11-28
GPT-2 transformer implementation

vankov:
## Description
This is a SparkNLP implementation of the GPT-2 model based on a TensorFlow graph exported from the HuggingFace GPT-2 implementation (TFGPT2LMHeadModel). The model supports is following closely the implementation of T5Transformer and has the same set of parameters. I had to modify the existing implementation of GPT2Tokenizer to align with the HF implementation. The main change was that I disabled the automatic prepending of a white space to each token - HF only does this for the token at the start of the text. I am not sure this is how it should work, but the results tend to be different, mainly because of appending a white space in front of every period, and I decided to stick to the HF code.

The notebook lets you export two TF servings but currently we only support the one without caching. I ran a number of tests on my local machine and found no advantage of caching. Need to investigate more, but it is possible that there is no performance gain of caching for this particular model.

## Motivation and Context

GPT-2 is a popular general-purpose model with multiple applications and would be nice to have it in SparkNLP.

## How Has This Been Tested?
There are several Scala tests and single Python one. I also tested exporting models from HF and loading them into SparkNLP. I've tried to run a test on Databricks but it takes ages to upload the jar and the model file. Will report if I manage to complete this test.

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x] My change requires a change to the documentation.


vankov:
The tests currently used a local copy of a GPT2 model. Once we have a pre-trained model uploaded to the model hub, the tests have to be updated.

DevinTDHa (approved):

maziyarpanahi (approved):

PR #6550, opened 2021-12-02
FEATURE NMH-51: Tracked latest model and change id for indexed models [skip test]

KshitizGIT:
Part of [NMH-51](https://johnsnowlabs.atlassian.net/browse/NMH-51)

- Only the latest models with same name and edition are indexed, thus removing duplicates during search

### DIFF
--- a/docs/_plugins/search_index.rb
+++ b/docs/_plugins/search_index.rb
@@ -2,6 +2,7 @@ require 'set'
 require 'uri'
 require 'net/http'
 require 'json'
+require 'date'
 require 'elasticsearch'
 require 'nokogiri'
 
@@ -132,9 +133,16 @@ end
 editions = Set.new
 uniq_to_models_mapping = {}
 uniq_for_indexing = Set.new
+name_language_editions_to_models_mapping = {}
 models_json = {}
 
 changed_filenames = []
+
+def is_latest?(group, model)
+  models = group[model[:id]]
+  Date.parse(model[:date]) == models.map { |m| Date.parse(m[:date])}.max
+end
+
 if File.exist?("./changes.txt")
   changed_filenames = File.open("./changes.txt") \
     .each_line \
@@ -181,7 +189,7 @@ Jekyll::Hooks.register :posts, :post_render do |post|
   end
 
   model = {
-    id: post.url,
+    id: "#{post.data['name']}_#{post.data['language']}_#{post.data['edition']}",
     name: post.data['name'],
     title: post.data['title'],
     tags_glued: post.data['tags'].join(' '),
@@ -192,7 +200,8 @@ Jekyll::Hooks.register :posts, :post_render do |post|
     edition_short: edition_short,
     date: post.data['date'].strftime('%F'),
     supported: !!post.data['supported'],
-    body: body
+    body: body,
+    url: post.url
   }
 
   uniq = "#{post.data['name']}_#{post.data['language']}"
@@ -201,6 +210,10 @@ Jekyll::Hooks.register :posts, :post_render do |post|
   editions.add(edition_short) unless edition_short.empty?
   uniq_for_indexing << uniq if changed_filenames.include?(post.basename)
   changed_filenames.delete(post.basename)
+
+  key = model[:id]
+  name_language_editions_to_models_mapping[key] = [] unless name_language_editions_to_models_mapping.has_key? key
+  name_language_editions_to_models_mapping[key] << model
 end
 
 client = nil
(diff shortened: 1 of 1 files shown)


KshitizGIT:
Request @pabla  for review

docs/_plugins/search_index.rb
```diff
@@ -192,10 +198,11 @@ def execute
     edition_short: edition_short,
     date: post.data['date'].strftime('%F'),
     supported: !!post.data['supported'],
-    body: body
+    body: body,
+    url: post.url
   }
 
-  uniq = "#{post.data['name']}_#{post.data['language']}"
+  uniq = "#{post.data['name']}_#{post.data['edition']}_#{post.data['language']}"
```
pabla, line 205: You break the purpose of this variable. It needs for collecting model editions. You need to create another variable with the same value.

docs/_plugins/search_index.rb
```diff
@@ -218,6 +225,18 @@ def execute
       },
     },
   )
+  exists =  client.indices.exists index: 'models'
```
pabla, line 228: Please don't create the index here. The index should contain more fields.

docs/_plugins/search_index.rb
```diff
@@ -249,8 +268,11 @@ def execute
 
       if client
         if force_reindex || uniq_for_indexing.include?(uniq)
-          id = model.delete(:id)
-          bulk_indexer.index(id, model)
+          if is_latest?(items, model)
+            id = model.delete(:id)
+            key = "#{model[:name]}_#{model[:edition]}"
```
pabla, line 273: Let's use id here.

docs/_plugins/search_index.rb
```diff
@@ -263,7 +285,7 @@ def execute
       filename.gsub! /\A(\d{4})-(\d{2})-(\d{2})-(\w+)\.md\z/, '/\1/\2/\3/\4.html'
     end.compact.each do |id|
       puts "Removing #{id}..."
-      client.delete index: 'models', id: id, ignore: [404]
+      client.delete_by_query index: 'models', body: {query: {term: {url: id}}}
```
pabla, line 288: I don't think we should change that.

pabla (changes requested):

docs/_plugins/search_index.rb
KshitizGIT: Currently, we are dynamically adding the fields.. This will only add the `keyword` type to the `url` field. When we bulk insert, all other fields will be created dynamically

docs/_plugins/search_index.rb
KshitizGIT: Agreed .. will change.

docs/_plugins/search_index.rb
pabla: Dynamic fields won't be created correctly.

pabla (changes requested):

docs/_plugins/search_index.rb
KshitizGIT: We need this, since we changed the _id in the indexed models

docs/_plugins/search_index.rb
KshitizGIT: Using `id`

docs/_plugins/search_index.rb
KshitizGIT: Adding the full mapping

docs/_plugins/search_index.rb
```diff
@@ -181,7 +189,7 @@ def execute
   end
 
   model = {
-    id: post.url,
+    id: "#{post.data['name']}_#{post.data['edition']}_#{post.data['language']}",
```
pabla, line 192: It is not really matter in this case, but we use a different order in other places, like model filenames:
- name
- language
- edition

I think it will be better for supporting in the future. Let's change the order.

docs/_plugins/search_index.rb
```diff
@@ -218,6 +231,58 @@ def execute
       },
     },
   )
+  exists =  client.indices.exists index: 'models'
+  puts "Index already exists : #{exists}"
+  unless exists
+    client.indices.create index: 'models', body: {
```
pabla, line 237: 👍

docs/_plugins/search_index.rb
```diff
@@ -218,6 +231,58 @@ def execute
       },
     },
   )
+  exists =  client.indices.exists index: 'models'
+  puts "Index already exists : #{exists}"
```
pabla, line 235: Please drop a space before colon.

pabla (changes requested):

docs/_plugins/search_index.rb
KshitizGIT: Done

docs/_plugins/search_index.rb
KshitizGIT: Done

docs/_plugins/search_index.rb
```diff
@@ -201,6 +210,10 @@ def execute
   editions.add(edition_short) unless edition_short.empty?
   uniq_for_indexing << uniq if changed_filenames.include?(post.basename)
   changed_filenames.delete(post.basename)
+
+  key = model[:id]
+  name_edition_language_models_group[key] = [] unless name_edition_language_models_group.has_key? key
```
pabla, line 215: Please rename this variable also.

pabla (changes requested):

docs/_plugins/search_index.rb
KshitizGIT: @pabla , Should I rename it to `name_language_editions_to_models_mapping` ?

pabla: Yes, something like that. I mean the `name_edition_language` part should be in the right order.

KshitizGIT: Done

pabla (approved):

maziyarpanahi (approved):

PR #6567, opened 2021-12-05
Use the best model while training NerDLApproach annotator 

maziyarpanahi:
This PR introduces `useBestModel` param in the NerDLApproach preserve and restore the model that has achieved the best performance at the end of the training. The priority is:
- metrics from `testDataset` param if it's set: will keep the model with the highest Micro F1 on test dataset
- metrics from `validationSplit` param if it's set: will keep the model with the highest Micro F1 on validation split
- if none is set it will keep track of `loss`: will keep the model with lowest loss


albertoandreottiATgmail:
@maziyarpanahi , I think you're missing the Python part.

maziyarpanahi:
@albertoandreottiATgmail I noticed that late and it was added in 340-release-candidate branch😊

src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorflowNer.scala
```diff
@@ -235,20 +250,50 @@ class TensorflowNer(val tensorflow: TensorflowWrapper,
...
     if (enableOutputLogs) {
       OutputHelper.exportLogFileToS3()
     }
+
+    lastCheckPoints
+
+  }
+
+  def saveBestModel(): Session = {
+    tensorflow.getSession()
```
albertoandreottiATgmail, line 296: @maziyarpanahi , are you sure this is going to freeze current state, and not just return a reference to a session that can be subsequently updated?

PR #6596, opened 2021-12-09
2021-12-09-bert_token_classifier_scandi_ner_xx

jsl-models:


josejuanmartinez (approved):
Looks good. Next time it's better to get real texts from Wikipedia. Thank you!

PR #6617, opened 2021-12-15
2021-12-15-mediacal_bert_token_classifier_ner_bacteria_en

jsl-models:


vkocaman:
@xusliebana now we have the medical bert4tokenCl !! great !

there is a typo in the name ; mediacal vs medical

PR #6641, opened 2021-12-20
2021-12-20-ner_drugprot_clinical_en

jsl-models:


luca-martial:
@vkocaman @muhammetsnts Murat checked the submission and asked me to assign you to review it. Thanks!

vkocaman:
@luca-martial can you explain what the entity types mean ? I see 4 classes in the original dataset but you have three over there. You probably merged GENE-N and GENE-Y but what is GENE_AND_CHEMICAL ?

https://zenodo.org/record/5119892#.YcBkpfEVA-Q

GENE-N  (non-normalizable mentions)
GENE-Y (normalizable to databases)
GENE
CHEMICAL

luca-martial:
Yes, so this decision was made with the advice of @josejuanmartinez

1. For condensing GENE-Y and GENE-N to GENE - this is because there are 3 datasets (train, development, test) and in the development and test sets GENE was not differentiated into GENE-Y and GENE-N so we decided to standardize and keep only GENE across all datasets

2. For GENE_AND_CHEMICAL - there was a significant proportion of chunks (~10%) that were labeled both as GENE and CHEMICAL (overlapping chunks/labels) - the authors of the dataset confirmed to us that this was on purpose since many of the tagged chemicals are also considered to be genes/proteins. So instead of separating these labels and training 2 different models, we decided to merge chunks with overlapping GENE and CHEMICAL labels to create a new label called GENE_AND_CHEMICAL

Let me know if this makes sense

murat-gunay:
@luca-martial 
Yes, there is a problem with overlapping entity labels.. there are two labels for the same chunks. but, This approach that you implemented may distort the original information of the dataset. We should look at customers' needs. if it fits with their needs and if it works and produce sensible insights then yes we can keep it like this.

luca-martial:
@murat-gunay that's a very good point

luca-martial:
Also for reference here is the corpus author's response about this issue @murat-gunay @vkocaman 

Each label or entity was annotated according to some well-defined rules (see annotation criteria below from the URLs). This also implies that there can be entity mentions of type GENE and of type CHEMICAL that overlap (this is especially true for enzymes) or there can be even identical (e.g. small peptides, according to rule "PEP").

See:
- https://zenodo.org/record/4957577#.Ybr1s33MKDU
- https://zenodo.org/record/4957519#.Ybr1tX3MKDU

We had chosen these annotation criteria to be able to capture relation types where the name of the GENE entity itself already has an implicit relation with a chemical (common in enzymes, transporters, or receptor genes/proteins). 

I think these aspects are captured by the "PEP" and the "M7 CEM overlapping with Enzymes" rules of the CHEMICAL entity annotation guidelines. You have to consider labels of GENE and CHEMICAL as independent entities.

josejuanmartinez:
Hi @vkocaman @luca-martial! Happy to see the model is trained. So, to summarize:

- GENE-Y and GENE-N were merged because there were datasets with GENE only, without the suffix. The dataset was not big, so I didn't want to discard anything.
- The authors decided to have overlapping labelled chunks GEN and CHEMICAL, that's why GEN_AND_CHEMICAL was created. We decided not to separate them and create different models because the overlapping is not huge, to keep 1 model, and not to affect much the Relation Extraction dataset we are going to use for training RE.

Cheers!

PR #6642, opened 2021-12-21
updated description with reasoning for entity labels

luca-martial:
Hi @vkocaman I've added explanations for entity labels in the description of the model. Let me know if those are enough or if I should elaborate. Thanks!


vkocaman:
looks good. thx !

PR #6648, opened 2021-12-21
2021-12-21-jsl_sbiobert_rxnorm_en

jsl-models:


vkocaman:
@galiph  the explanation you add looks not good.. let sue the following instead.

'This model maps sentences & documents to a 768 dimensional dense vector space by using average pooling on top of BioBert model. It's also fine tuned on RxNorm dataset to help generalisation over medication-related dataset.'

vkocaman:
@galiph we need to rename this model to 'jsl_sbiobert_cased_rxnorm_en'
(it would be probably better to close this PR and open a new one due to possible conflicts in model hub)

PR #6650, opened 2021-12-22
2021-12-22-sbertresolve_jsl_rxnorm_augmented_med_en

jsl-models:


vkocaman:
@galiph  the examples are wrong.. you feed the entire sentcne and expect the model to return the rxnorm codes. This is a shot pipeline you sue dover there.. so you need to use the right pipeline

vkocaman:
@galiph  the examples all look like 1:1 mapping.. it would be better if we can add other examples that the model can find  by contextual similarity (T2DM --> type 2 diabetes etc)

PR #6695, opened 2022-01-03
2022-01-03-bert_token_classifier_ner_bionlp_en

jsl-models:


vkocaman:
@egenc this is a wrong one.. it should be medicalBert4token

in the next PRs, please assign @josejuanmartinez as a reviewer at first

PR #6705, opened 2022-01-04
2022-01-04-bert_token_classifier_ner_ade_en

jsl-models:


vkocaman:
@egenc lets add `.setMaxSentenceLength(512)` in each example for this annotator (as we are not using Sent Detector at all)

egenc:
> @egenc lets add `.setMaxSentenceLength(512)` in each example for this annotator (as we are not using Sent Detector at all)

I have done it @vkocaman , I can try it once you merge

PR #6709, opened 2022-01-04
2022-01-04-check_spelling_dl_en

jsl-models:


maziyarpanahi (approved):

maziyarpanahi (approved):

josejuanmartinez (changes requested):
The scala and the python examples are different, and the pretrained pipelines too (clean_slang vs check_spelling_dl"). This is the check_spelling_dl markdown, so probably Python section is the wrong one?

docs/_posts/egenc/2022-01-04-check_spelling_dl_en.md
```diff
@@ -0,0 +1,104 @@
...
+<div class="tabs-box" markdown="1">
+{% include programmingLanguageSelectScalaPythonNLU.html %}
+```python
+pipeline_local = PretrainedPipeline("check_spelling_dl")
+
+testDoc = '''
+During the summer we have the hottest ueather. I have a black ueather jacket, so nice.I intrduce you to my sister, she is called ueather.
+'''
+
+result=pipeline_local.annotate(testDoc)
```
josejuanmartinez, line 40: We are having fullAnnotate in scala, and annotate in Python. I think the results you are showing are from annotate, since it's free text and not Annotation objects. Right?

josejuanmartinez (changes requested):
I think there is still an issue with annotate and fullAnnotate. Take a look at the comment.

PR #6741, opened 2022-01-07
Revert back to none optional session for more reliable serializations

maziyarpanahi:
If Java serialization is not set to Kyro or sometimes for a random reason user may get `NullPointerException` in TensorflowWrapper when we use `Session` as Optional.


maziyarpanahi:
[Revert tf session Optinal type](https://trello.com/c/BFxg29Fx/497-revert-tf-session-optinal-type)

PR #6773, opened 2022-01-15
Implement TF Session warmup for some TF-based annotators

maziyarpanahi:
When we create a TF session for the first time, the time for inferencing for some TF models is slower the first time! TensorFlow recommends warmup for some models, an empty pass over the session which we can do when we initiate the TensorFlow class with empty values.

Ref: https://www.tensorflow.org/tfx/serving/saved_model_warmup

> The TensorFlow runtime has components that are lazily initialized, which can cause high latency for the first request/s sent to a model after it is loaded. This latency can be several orders of magnitude higher than that of a single inference request.

> To reduce the impact of lazy initialization on request latency, it's possible to trigger the initialization of the sub-systems and components at model load time by providing a sample set of inference requests along with the SavedModel. This process is known as "warming up" the model.

Architecture:

- [x] MarianTransformer
- [x] T5Transformer
- [x] GPT2Transformer

Doing the warmup when initializing these annotators will shave off 10s-15s from the first inference (action in Spark such as .count, .show, .save/.write, etc.)

Also includes: https://github.com/JohnSnowLabs/spark-nlp/pull/6781


maziyarpanahi:
[Find a solution for TF Warmup](https://trello.com/c/VVU8hxcC/444-find-a-solution-for-tf-warmup)

PR #6825, opened 2022-01-26
2022-01-26-word2vec_wiki_1000_fr

jsl-models:


josejuanmartinez (changes requested):
There is an example of how Scala code shoud look like in the Models Hub channel. Take a look at the absence of backslashesh, the usage of Array(), etc.

Also, this kind person asked to cite him this way.. You can use that in References section.

image

luca-martial:
@josejuanmartinez let me know if the edits are okay?

maziyarpanahi:
@luca-martial should we close this in favor of the new embeddings just uploaded?

luca-martial:
@maziyarpanahi I guess I would leave that call up to you, these ones are based on the 
FrWiki dump whereas the other ones are based on the [frWaC corpus](https://wacky.sslmit.unibo.it/doku.php?id=corpora). this one was trained on 600 million words whereas the other one (frWaC) was trained on 1.6 billion so there is a significant difference in performance (frWaC is much better)

maziyarpanahi:
> @maziyarpanahi I guess I would leave that call up to you, these ones are based on the
> FrWiki dump whereas the other ones are based on the [frWaC corpus](https://wacky.sslmit.unibo.it/doku.php?id=corpora). this one was trained on 600 million words whereas the other one (frWaC) was trained on 1.6 billion so there is a significant difference in performance (frWaC is much better) - I guess users would prefer the better one?

I understand better now. The other one is better because it’s bigger. We will leave both of them and users can choose depending on the resources they have.

luca-martial:
yes in terms of words used, but actually the dimensions of the embeddings from the better performing one (frWaC) is 200, whereas this one (frwiki) is 1000, so the size of frwiki is larger (253mb vs 120mb)

luca-martial:
there's no real advantage of frwiki except from the fact that it was trained on wikipedia dumps... which im not sure how important that is to users

maziyarpanahi:
> there's no real advantage of frwiki except from the fact that it was trained on wikipedia dumps... which im not sure how important that is to users

Let's keep both, Wiki dumps are more generic in terms of domains and might perform better on different datasets. (it's good to have the options especially on none-English languages where there are not that many choices)

I think @josejuanmartinez will merge this soon

maziyarpanahi (approved):

josejuanmartinez (changes requested):
An Array() is Scala is required when bracks in Python are. Can you fix it in the last component in your pipeline?

luca-martial:
@josejuanmartinez fixed!

josejuanmartinez (approved):
Looks good now, thank you!

PR #6842, opened 2022-02-01
2022-02-01-ner_deid_subentity_fr

jsl-models:


luca-martial:
hey @josejuanmartinez @vkocaman i'm uploading new word embeddings, so the sample code for this model contains what I think will be the name of the word embeddings, but it's not 100% sure yet: 
```
embeddings = WordEmbeddingsModel.pretrained("word2vec_wac_200", "fr")\
	.setInputCols(["sentence", "token"])\
	.setOutputCol("word_embeddings")
```

luca-martial:
These were the [old embeddings](https://github.com/JohnSnowLabs/spark-nlp/pull/6825) - not good. These are the [new embeddings](https://github.com/JohnSnowLabs/spark-nlp/pull/6843) - much better.

josejuanmartinez (changes requested):
Hi! Good job. In scala code, you are missing the Array() constructor where an array is expected. To see when it is, check brackets in Python code.

Also, although I did not that in the Spanish one, I think using a multilingual sentence detector ('xx') makes more sense.

luca-martial:
@josejuanmartinez fixed!

luca-martial:
uploading new, better performing version

PR #6843, opened 2022-02-01
2022-02-01-word2vec_wac_200_fr

jsl-models:


vkocaman:
It makes sense to use Fr embeddings for any clinical french model.. I wish we could find one trained on clinical french corpora but looks like there is no such (worth to look around)

can you check the word coverage of this model on french Deid conll ?

PR #6845, opened 2022-02-02
Fix GraphExtraction crashing when fullAnnotate and document are used at the same time

maziyarpanahi:
This PR fixes an exception that happens during the following scenario:

- One of the inputCols in GraphExtraction is `document` instead of `sentence`
- `.fullAnnotate` is used from `LightPipeline`

Then it crashes due to not having a `sentence` index inside metadata coming from `.fullAnnotate`


danilojsl (approved):
LGFM

PR #6867, opened 2022-02-05
Fix broken BertEmbeddings wrong sentence indexing

maziyarpanahi:
In Spark NLP 3.3.3 release we introduced a bug that uses a rowIndex for sentence's index rather than preserving an already existing sentence's index in metadata. This PR fixes this bug.


maziyarpanahi:
[BertEmbeddings overrides sentenceIndex from 0 to batchSize](https://trello.com/c/FKfUKyCu/523-bertembeddings-overrides-sentenceindex-from-0-to-batchsize)

PR #6868, opened 2022-02-05
Fix broken setExceptionsPath in Tokenizer in Python

maziyarpanahi:
This PR fixes the wrong type of exceptionsPath param and add a setter to Tokenizer in Python APIs.


maziyarpanahi:
[setExceptionsPath is missing from Tokenizer Python APIs](https://trello.com/c/NWWsfULd/524-setexceptionspath-is-missing-from-tokenizer-python-apis)

PR #6904, opened 2022-02-11
2022-02-11-distilbert_base_sequence_classifier_qqp_en

jsl-models:


josejuanmartinez (changes requested):
Duplicate**d** Question Detection.

Also, David asked not to include LABEL_0 and LABEL_1 labels. When importing from HF, please change the labels.txt in the metadata, so that we show something meaningful ("duplicated", "not_duplicated")

murat-gunay:
> > Duplicate**d** Question Detection.
> > Also, David asked not to include LABEL_0 and LABEL_1 labels. When importing from HF, please change the labels.txt in the metadata, so that we show something meaningful ("duplicated", "not_duplicated")
> 
> Changing `labels.txt` file before importing to Spark NLP should be the easiest and safest. (if you do modify metadata it has a .crc hidden file to check for corruption, that has to be removed or it won't load a modified file)

It seems I will re-import  and re-upload the model again...

PR #6957, opened 2022-02-22
2022-02-22-bert_sequence_classifier_antisemitism_pipeline_en

jsl-models:


murat-gunay:
I am not sure if `new` may not be needed in scala script.

murat-gunay:
image

josejuanmartinez:
> I am not sure if `new` may not be needed in scala script.

It's needed, thank you!

josejuanmartinez (changes requested):
I know it's late to change `bert_sequence_classifier_antisemitism_en` but for the pipeline, could we return something not 0 / 1 ? We should not have numerical outputs, always classes. In this case, True or False.

murat-gunay:
> I know it's late to change `bert_sequence_classifier_antisemitism_en` but for the pipeline, could we return something not 0 / 1 ? We should not have numerical outputs, always classes. In this case, True or False.

there is no way other than re-importing the model again from HF, I guess.

josejuanmartinez (approved):

PR #6962, opened 2022-02-23
2022-02-23-clinical_deidentification_fr

jsl-models:


luca-martial:
> Probably we should obfuscate SEX just as "Homme" or "Femme", because of this:
> 
> image

@josejuanmartinez so just 1 of "Homme" or "Femme" or both options "Homme" and "Femme"?

josejuanmartinez:
Both

luca-martial:
new model uploaded

PR #7170, opened 2022-03-14
2022-03-14-bert_token_classifier_ner_chemicals_pipeline_en_2_4

jsl-models:


murat-gunay:
This is pyspark 2.4 version of the `bert_token_classifier_ner_chemicals_pipeline_en` pipeline

murat-gunay (approved):

vkocaman:
@murat-gunay why do we need pretrained pipelines for NERs ? we usually do that for the pipelines having multiple models from different annotators (NER + something -- RE, ER etc)

PR #7455, opened 2022-03-21
2022-03-21-bert_token_classifier_ner_jsl_en_2_4

jsl-models:


muhammetsnts (commented):
Versions had to be the same versions with the broken model. I edited all version cells and the model path.

muhammetsnts (approved):

PR #7486, opened 2022-03-29
Docs: Update SentenceDetector

DevinTDHa:
## Description

This PR updates documentation for the SentenceDetector to make it more transparent for the user. 

@maziyarpanahi You mentioned in the issue, that the user should be able to replicate the rules themselves. Is this [high-level description](https://github.com/JohnSnowLabs/spark-nlp/pull/7486/commits/3e89172134ab613a74c651acd45477096cc2e307#diff-79e3b9f663e2b5195f0f2c20f9867b6e2839bcfb244a06463f7736ba93707e93R6-R23) enough or should it be more explicit?

## How Has This Been Tested?

No code changes.

### DIFF
--- a/docs/en/annotator_entries/SentenceDetector.md
+++ b/docs/en/annotator_entries/SentenceDetector.md
@@ -3,7 +3,35 @@ SentenceDetector
 {%- endcapture -%}
 
 {%- capture description -%}
-Annotator that detects sentence boundaries using any provided approach.
+Annotator that detects sentence boundaries using regular expressions.
+
+The following characters are checked as sentence boundaries:
+
+1. Lists ("(i), (ii)", "(a), (b)", "1., 2.")
+2. Numbers
+3. Abbreviations
+4. Punctuations
+5. Multiple Periods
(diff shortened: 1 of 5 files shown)


maziyarpanahi:
@DevinTDHa The explanation is great, but the example is very simple/easy compared to what we use as default in `extractBounds`. So if someone wants everything we have but not the `;`, there is no option to replicate that. (can we have an example as close as possible to what we have inside `extractBounds`? I know it might be very massive, but just a few from each rule)

maziyarpanahi (approved):

DevinTDHa:
Hi @maziyarpanahi, 

For such an example, do you think it'd be better to create a spark-nlp-workshop notebook instead and link it as an advanced example? Or would you like it inside the documentation?

maziyarpanahi:
@DevinTDHa Yes, definitely a notebook with different examples from simple to complex (close to the default rules) would be the best. We can leave a link to that notebook in scaladoc/pydoc

DevinTDHa:
I have added the new workshop notebook here:

https://github.com/JohnSnowLabs/spark-nlp-workshop/pull/515

PR #7669, opened 2022-04-04
2022-04-04-drug_action_treatment_mapper_en_3_0

jsl-models:


Ahmetemintek:
We've uploaded this model with v3.5.0 but it returned an error. This time we upload it with v3.4.2 by the suggestion of Jesus.

PR #7685, opened 2022-04-07
impossiblePenultimates added to SentenceDetectorDLModel

vankov:
Add impossiblePenultimates parameter to the model.

## Description
The parameter was only available for the Approach, not it is also usable in the model. 

Note that removing some of the existing impossible penultimates (i.e. the ones used in the approach for training the model) may not necessary result in splitting the sentence around the penultimate. For example, if the penultimate "dr" was used in the approach, then the approach have never encountered any sentence ending with "dr." and therefore just removing the penultimate doesn't not necessary entail the the model will start splitting sentences around "dr." 

Adding new penultimates always works.

## Motivation and Context

Users have requested the ability to modify/extend the list of impossible penultimates of already trained models.

## How Has This Been Tested?

The changes have been test in both Scala and Python. The tests appended a new penultimate and verified no sentences ending with it are detected.

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x] My change requires a change to the documentation.

### DIFF
--- a/python/sparknlp/annotator.py
+++ b/python/sparknlp/annotator.py
@@ -10948,6 +10948,9 @@ class SentenceDetectorDLModel(AnnotatorModel):
         Set the minimum allowed length for each sentence, by default 0
     maxLength
         Set the maximum allowed length for each sentence, by default 99999
+    impossiblePenultimates
+        Impossible penultimates - list of strings which a sentence can't end
+        with
 
     Examples
     --------
@@ -11032,6 +11035,11 @@ class SentenceDetectorDLModel(AnnotatorModel):
                       "Set the maximum allowed length for each sentence",
                       typeConverter=TypeConverters.toInt)
 
+    impossiblePenultimates = Param(Params._dummy(),
+                                   "impossiblePenultimates",
+                                   "Impossible penultimates - list of strings which a sentence can't end with",
+                                   typeConverter=TypeConverters.toListString)
+
     def setModel(self, modelArchitecture):
         """Sets the Model architecture. Currently only ``"cnn"`` is available.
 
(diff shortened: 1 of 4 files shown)


maziyarpanahi (approved):

jasonlilley:
Hello, I have questions about this feature that I asked in the SparkNLP Slack channel, but didn't get any useful response. I want to use these models for a project. I was under the impression that if I gave .setImpossiblePenultimates() a list of strings, it would stop splitting sentences in a way that the string would end the sentence. In my own data, I would prefer that the model not end a sentence after a colon (":"). So I tried defining my pipeline as follows:

> documentAssembler2 = DocumentAssembler() \
>     .setInputCol("note") \
>     .setOutputCol("document")
> 
> sentenceDL2 = SentenceDetectorDLModel\
>     .pretrained("sentence_detector_dl","en") \
>     .setInputCols(["document"]) \
>     .setOutputCol("sentence") \
>     .setImpossiblePenultimates([":"])
> 
> pipeDL2 = Pipeline().setStages([
>     documentAssembler2,
>     sentenceDL2
> ])
> 
> mydfDL2 = pipeDL2.fit(mydf).transform(mydf).collect()

I ran this, and then compared it to the output of an identical pipeline without the .setImpossiblePenultimates([":"]) line. The output is different, but in ways that seem to have nothing to do with colons (even documents without colons were different). I still have sentences that end with ":" as before. In fact, rather than fewer sentences, the output now has more sentences. It now seems to break more often after periods -- maybe after particular abbreviations in particular, e.g. "Dr." and "pm". I wonder if by trying to add a new string, it is losing the default string list it was using before. But I don't know where to find the default list so that I can add it to my code. Also, it is still splitting after colons. Does the method only work for alphabetic strings?

PR #7881, opened 2022-04-11
Tokenizer: Optimized tokenization with exceptions

DevinTDHa:
## Motivation and Context

The current Tokenizer implementation has performance issues if a large number of exceptions are used. 

For example:
When tokenizing `src/test/resources/spell/sherlockholmes.txt`, on my machine it will take an average of 0.58 seconds for 20 iterations. If there are 100 exceptions, it will take on average 3.17 seconds, for 200 it will jump to 5.8 seconds. The following benchmark was used for testing: [TokenizerTestSpec](https://github.com/DevinTDHa/spark-nlp/blob/c2506ceb90ee33998e02449b9f6f623523972092/src/test/scala/com/johnsnowlabs/nlp/annotators/TokenizerTestSpec.scala#L665)

There is double the compilations of regular expressions and a severe performance drop. 

### Problem

The issue lies with how the exceptions are checked. 
1. Regex are compiled for each sentence that is tagged. This can be done beforehand.
    - Addtionally, each regex exception is compiled twice due to first being compiled to replace the `BREAK_PATTERN` in [`tag`](https://github.com/JohnSnowLabs/spark-nlp/blob/24524a7fbb73fb2411879e062fd33b89b3dddbb1/src/main/scala/com/johnsnowlabs/nlp/annotators/TokenizerModel.scala#L244) and afterwards compiled again in [`casedMatchExists`](https://github.com/JohnSnowLabs/spark-nlp/blob/24524a7fbb73fb2411879e062fd33b89b3dddbb1/src/main/scala/com/johnsnowlabs/nlp/annotators/TokenizerModel.scala#L226-L228)

## Description

The following changes were implemented:
  - exceptions and regular expressions are compiled during instantiation of the model instead of for each sentence
  - exceptions are combined into one regular expression in [`TokenizerModel`](https://github.com/DevinTDHa/spark-nlp/blob/c2506ceb90ee33998e02449b9f6f623523972092/src/main/scala/com/johnsnowlabs/nlp/annotators/TokenizerModel.scala#L262-L277)
  - function `casedMatchExists` was removed in favor of a hash set that checks whether an exception without a break exists in [`tag`](https://github.com/DevinTDHa/spark-nlp/blob/c2506ceb90ee33998e02449b9f6f623523972092/src/main/scala/com/johnsnowlabs/nlp/annotators/TokenizerModel.scala#L292-L313)
    - @maziyarpanahi Do you think this approach is acceptable?
  - fixed some minor warnings highlighted by intelliJ

### Benchmark Results
These changes will result in a performance boost of 
  - more than 300% for 100 exceptions (from 3.17 down to 0.95 seconds) and 
  - for 200 exceptions a performance boost of more than 450% (from 5.8 seconds down to 1.24 seconds).

## How Has This Been Tested?

Added an additional test for the benchmark. All Tokenizer tests are passing.

## Types of changes

## Checklist:

### DIFF
--- a/src/main/scala/com/johnsnowlabs/nlp/annotators/TokenizerModel.scala
+++ b/src/main/scala/com/johnsnowlabs/nlp/annotators/TokenizerModel.scala
@@ -16,12 +16,15 @@
 
 package com.johnsnowlabs.nlp.annotators
 
+import com.johnsnowlabs.nlp._
 import com.johnsnowlabs.nlp.annotators.common._
 import com.johnsnowlabs.nlp.serialization.StructFeature
 import com.johnsnowlabs.nlp.util.regex.RuleFactory
-import com.johnsnowlabs.nlp._
 import org.apache.spark.ml.param.{BooleanParam, IntParam, Param, StringArrayParam}
 import org.apache.spark.ml.util.Identifiable
+
+import scala.collection.mutable
+import scala.util.matching.Regex
 
 /** Tokenizes raw text into word pieces, tokens. Identifies tokens with tokenization open
   * standards. A few rules will help customizing it if defaults do not fit user needs.
@@ -68,7 +71,7 @@ class TokenizerModel(override val uid: String)
   val exceptions: StringArrayParam =
     new StringArrayParam(this, "exceptions", "Words that won't be affected by tokenization rules")
 
-  /** Whether to care for case sensitiveness in exceptions
+  /** Whether to care for case sensitiveness in exceptions (Default: `true`)
     *
     * @group param
     */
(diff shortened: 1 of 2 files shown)


DevinTDHa:
@maziyarpanahi I will test how well it performs. In this particular case, the hashset will not grow much: at most it will have nr. of exceptions entries and only if these exceptions are not affected by the `BREAK_PATTERN`.

DevinTDHa:
@maziyarpanahi I looked at it the following ways:

I used free samples form the COCA Dataset as a test. In total these are about 500k sentences, 56 million characters.

As exceptions I chose something ridiculous like `\w+`, which will effectively match every word in a sentence. I also monitored the tests with visualvm.

- Tokenizing it the reuglar way has no issues. The set doesn't grow beyond the vocabulary of the sentence and the set also only exists for each sentence, keeping it small.
- For the second test, I replaced all `\n` with `\t` in the sample data. This effectively results in one massive sentence, which will be tokenized at once. The collection will grow to a fixed size of the vocabulary of the data (which only happens once). The processing of the exceptions will be done in about 5 seconds on my machine. 
  - **Note:** With the current Tokenizer Implementation, the whole tokenization process doesn't terminate (at least I didn't wait for that long). Followingly, the current implementation of the exceptions will perform even worse.

maziyarpanahi (approved):

PR #7933, opened 2022-04-14
2022-04-14-spellcheck_clinical_en_2_4

jsl-models:


murat-gunay:
This is for the PySpark 2.4.4 with SparkNLP 3.4.2 and above users.

josejuanmartinez:
@murat-gunay There are conflicts to resolve

murat-gunay:
The conflict is probably because of the same name/data/title of the models.. The simple way is to close this PR and try to re-upload on a different day.

PR #7934, opened 2022-04-14
2022-04-14-spellcheck_clinical_pipeline_en_2_4

jsl-models:


murat-gunay:
This pipeline is for PySpark 2.4.x users with SparkNLP 3.4.1.

PR #8015, opened 2022-04-14
2022-04-14-albert_embeddings_albert_german_ner_de_3_0

jsl-models:


docs/_posts/luca-martial/2022-04-14-albert_embeddings_albert_german_ner_de_3_0.md
```diff
@@ -0,0 +1,90 @@
...
+## Model Information
+
+{:.table-model}
+|---|---|
+|Model Name:|albert_embeddings_albert_german_ner|
+|Compatibility:|Spark NLP 3.4.2+|
+|License:|Open Source|
+|Edition:|Official|
+|Input Labels:|[sentence, token]|
+|Output Labels:|[bert]|
```
maziyarpanahi, line 83: when uploading models, the `setOutputCol` will be stored inside the metadata of the model. That is why here it says `bert` which should be either `embeddings` or `albert`.

luca-martial: @maziyarpanahi thanks for the feedback, in terms of fix would you suggest closing all the PRs and re-uploading all the models with fixed metadata? Or is there a more efficient way around this?

PR #8046, opened 2022-04-18
2022-04-18-spellcheck_clinical_en_2_4

jsl-models:


murat-gunay:
This model is trained for PySpark 2.4.x users with SparkNLP 3.4.2 and above.

josejuanmartinez (approved):

PR #8069, opened 2022-04-22
2022-04-22-bert_ner_bert_base_chinese_pos_zh_3_0

jsl-models:


gadde5300:
Closed because the tags were not added.

PR #8322, opened 2022-05-09
2022-05-09-rxnorm_ndc_mapper_en_3_0

jsl-models:


vkocaman:
@galiph product and package NDCs are always same ? I suppose not. Can you use an example that returns different  product and package codes ?

PR #8410, opened 2022-05-11
2022-05-11-drug_brandname_ndc_mapper_en_3_0

jsl-models:


vkocaman:
@galiph looks like there is an issue with scala example
image

and why do we load all from local ? it should be `ChunkMapperModel.pretrained(xxx)`

vkocaman:
thx !

galiph:
You are right. I have just corrected it.

PR #8505, opened 2022-05-17
Enabling in-memory param for WordEmeddings

danilojsl:
## Description

This change introduces a new parameter `enableInMemoryStorage` that allows the user to define the embeddings lookup strategy. 

- When `setEnableInMemoryStorage = True`, means that the embeddings lookup will be in-memory. This will be faster than using RocksDB but at the cost of consuming more memory.
- By default, it is set to `False`, which means it will use the disk to store and lookup embeddings through RocksDB component.

This change only works for `WordEmbeddings` annotator. However, the changes to `HasStorage` and `HasStorageModel` interfaces make future changes open to other annotators that use RocksDB feature.

This change also removes `useStorage` parameter from `EntityRuler` annotator, since `enableInMemoryStorage` will handle such feature

## Motivation and Context

- Enabling using a normal in-memory lookup to solve S3/Kubernetes issues where there is no distributed supported fileSystem
- Faster processing of WordEmbeddings since it will use full in-memory lookup

## How Has This Been Tested?

Unit tests
Google Colab notebook

## Types of changes

## Checklist:


maziyarpanahi (approved):

danilojsl:
Closed since a new PR with a different design will be created for this feature

PR #8533, opened 2022-05-20
M1 support, upgrade to newer RocksDB and AWS Java SDK

DevinTDHa:
## Description

This PR enables Spark NLP to be built with M1 support. Most of the required packages have already been published and have been integrated into the build.

This PR fixes the remaining compatibility issues:

- Add another condition in `build.sbt` to use a newer `rocksdbjni` version if m1 was selected for the build.
- Add compatible libraries for libsentencepiece_jni and _sentencepiece_tokenizer with respective conditions for loading

TODO:

- [ ]  Update docs with user guide on how to use m1 jar.

## Types of changes


maziyarpanahi:
@DevinTDHa Can we not call those native files `aarch64`? (I am not sure if these files are compatible with Linux aarch64 which Databricks/AWS are offering now, so when we start supporting those it maybe confusing - but if they are compatible with all `aarch64` and not just M1 we can leave it like this, otherwise let's replace `aarch64` with `m1`)

DevinTDHa:
> @DevinTDHa Can we not call those native files `aarch64`? (I am not sure if these files are compatible with Linux aarch64 which Databricks/AWS are offering now, so when we start supporting those it maybe confusing - but if they are compatible with all `aarch64` and not just M1 we can leave it like this, otherwise let's replace `aarch64` with `m1`)

Yes, let me change it actually to M1 so it is clearer. Regarding Linux and Windows compatibility there shouldn't be an issue, as this is for the mac dylibs only (i'll add another condition for `SentencePieceJNI.java` though)

maziyarpanahi (approved):

PR #8559, opened 2022-05-27
FEATURE NMH-94: Add model naming conventions [skip test]

pabla:
Add a toggle button for model naming conventions.

model_naming_conventions

### DIFF
--- a/docs/_frontend/package.json
+++ b/docs/_frontend/package.json
@@ -24,6 +24,8 @@
     "webpack-dev-server": "^4.7.4"
   },
   "dependencies": {
+    "@floating-ui/react-dom": "^0.7.1",
+    "@floating-ui/react-dom-interactions": "^0.6.3",
     "classnames": "^2.3.1",
     "downshift": "^6.1.7",
     "preact": "^10.7.0"
--- a/docs/_frontend/src/models/App/App.js
+++ b/docs/_frontend/src/models/App/App.js
@@ -1,8 +1,7 @@
+import React, { useState, useEffect } from 'react';
 import Form from '../Form';
 import ModelItem from '../ModelItem';
 import ModelItemList from '../ModelItemList';
-
-const { createElement: e, useState, useEffect } = React;
 
 const SEARCH_ORIGIN = 'https://search.modelshub.johnsnowlabs.com';
 
@@ -53,6 +52,7 @@ const fromSearchString = () => {
     'tags',
     'sort',
     'predicted_entities',
+    'page',
   ].forEach((key) => {
     if (!searchParams.has(key)) {
       return;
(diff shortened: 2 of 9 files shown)


josejuanmartinez:
The only thing I'm thinking about right now, is - that information is very Healthcare related. Maybe we can have also other naming conventions from Open Source and OCR.

maziyarpanahi:
@pabla Are the options in that box dynamic? (like if there is no BioBERT among the results, it won't show it right?)

josejuanmartinez:
> @pabla Are the options in that box dynamic? (like if there is no BioBERT among the results, it won't show it right?)

I don't think they are. What I don't know is how difficult it is to do that. But would be nice. Because right now all we have is healthcare related.

maziyarpanahi:
> > @pabla Are the options in that box dynamic? (like if there is no BioBERT among the results, it won't show it right?)
> 
> I don't think they are. What I don't know is how difficult it is to do that. But would be nice. Because right now all we have is healthcare related.

`large, `medium`, `small`, and `fast` can be public related. We have lots of models/pipelines with those names in their descriptions/names. But I hope we can update that list based on the results so we don't have to show a large list of names every time. (but we can have a place as `how to use Models Hub` to have all these info)

josejuanmartinez:
thank you for the info about the open source models conventions. Let's have
it in the same models hub page, since this came from several clients
requests and Pavel has done already much work there. And in them meantime
for future updates we can think if we want them somewhere else.

maziyarpanahi (approved):

pabla:
@maziyarpanahi No, the options aren't dynamic right now, but it is a good idea, let's implement that.

pabla:
@maziyarpanahi @josejuanmartinez 
Maybe then is it worth to highlight naming conventions in the model titles and display them on hover instead of a button in the header?

Screenshot 2022-05-30 at 12 07 59 pxd copy@2x

josejuanmartinez:
Yes, let's hover them, and also include the conventions that Maz commented
above

pabla:
> Yes, let's hover them, and also include the conventions that Maz commented above.

@josejuanmartinez 
Could you please provide me descriptions for those conventions? (medium, small, fast)

josejuanmartinez:
@maziyarpanahi can you provide a short desc for the open source conventions?
large, medium, small, and fast

maziyarpanahi:
> @maziyarpanahi can you provide a short desc for the open source conventions? large, medium, small, and fast

sure:

large: used the largest size Word/Sentence Embeddings in this model/pipeline
medium: used the average size Word/Sentence Embeddings in this model/pipeline
small: used the smallest size Word/Sentence Embeddings in this model/pipeline
fast: used the fastest algorithms in this model/pipeline (speed over accuracy)

pabla:
@josejuanmartinez @maziyarpanahi 
Here is an updated UI:
https://user-images.githubusercontent.com/1594817/171149292-442cc452-1e5e-4c4d-b796-d4c0c9a44202.mov

josejuanmartinez:
@pabla  this looks great. I took a look to look for more naming conventions. Can you add the following?
- cased: This model differentiates between lowercase and uppercase.
- uncased: This model does not differentiate between lowercase and uppercase.
- normalized: in-house standardization and normalization of input and output labels

Also, I found some titles are not being fully shown, that will prevent to show the naming conventions in the title if they are at the end, right?
image

pabla:
> Also, I found some titles are not being fully shown, that will prevent to show the naming conventions in the title if they are at the end, right?

Currently it works as you said. But after applying this PR, the title will be displayed fully (in the video above, the third card has a title with 5 lines).

josejuanmartinez (approved):

PR #9101, opened 2022-06-08
2022-06-08-distilbert_qa_AQG_CV_Squad_en_3_0

jsl-models:
@josejuanmartinez 
Please take a look at this initial QA-DistilBert model for an example. I will upload 165 DistilBertQA models after your "OK".


murat-gunay:
I checked all "Model Titles" and fixed manually if there was duplicates.. There are 165 unique Model Name and Model Title in this batch.

PR #9263, opened 2022-06-11
Feature/mahmood add an option for displaying recommended models first nmh 97

mahmoodbayeshi:
related to [NMH-97](https://johnsnowlabs.atlassian.net/browse/NMH-97)

Screenshot 2022-06-15 at 11 35 01

### DIFF
--- a/docs/_frontend/src/models/App/App.js
+++ b/docs/_frontend/src/models/App/App.js
@@ -27,10 +27,14 @@ const toSearchString = (params) => {
             acc.append(k, params[k]);
           }
           break;
+        case 'recommended':
+          break;
         default:
           acc.append(k, params[k]);
           break;
       }
+    } else {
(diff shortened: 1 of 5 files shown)


docs/_frontend/src/models/App/App.js
```diff
@@ -64,7 +68,7 @@ const fromSearchString = () => {
         break;
 
       case 'supported':
-        if (params[key] === '1') {
+        if (searchParams.get(key) === '1') {
```
mahmoodbayeshi, line 71: During the test, I noticed this bug that caused the support checkbox doesn't update from URL params so I fixed it,

pabla (commented):
Mahmood, please change the target branch (it should be `master`).

pabla (approved):

PR #9327, opened 2022-06-18
2022-06-18-bert_token_classifier_drug_development_trials_en_3_0

jsl-models:


vkocaman:
@luca-martial this might be helpful for clinical trials IP

PR #9565, opened 2022-06-21
FEATURE NMH-96: Remove models with changes to Elastic search key [skip-test]

KshitizGIT:
Fix for Jira Ticket [NMH-96](https://johnsnowlabs.atlassian.net/browse/NMH-96)


docs/_plugins/search_index.rb
```diff
@@ -439,6 +432,20 @@ def is_latest?(group, model)
   end
   bulk_indexer.execute
 
+  # remove renamed or deleted posts from the index
+  if client
+    changed_urls = changed_filenames.map do |filename|
+      filename.gsub! /\A(\d{4})-(\d{2})-(\d{2})-(\w+)\.md\z/, '/\1/\2/\3/\4.html'
+    end.compact
+    unless changed_urls.empty?
+      client.delete_by_query index: 'models', body: {query: {bool: {must: {terms: {url: changed_urls}}}}}
```
pabla, line 441: You remove all documents below, and this request looks useless.

KshitizGIT: Hmm, I just reverted back to the old state. This case handles when we rename the file path of the model (i.e the url changes for the model but the id remains same). I kept it just to be sure that changing the file path also deletes older documents.
I guess we can remove it as in Elastic Search you can't have two docs with the same id.

pabla (approved):

maziyarpanahi (approved):

PR #9579, opened 2022-06-21
2022-06-21-ner_trials_abstracts_en_3_0

jsl-models:


luca-martial:
@vkocaman is the model name okay?

vkocaman:
@luca-martial did you train a final model with no val split ? model name would be `ner_clinical_trials_abstracts`

luca-martial:
> @luca-martial did you train a final model with no val split ? model name would be `ner_clinical_trials_abstracts`

yes, i did - i will change the model name and reupload

PR #9591, opened 2022-06-22
2022-06-22-ner_clinical_trials_abstracts_en_3_0

jsl-models:


luca-martial:
reuploaded with version label fixes - good to merge

luca-martial:
@vkocaman this is ready to merge, the model renaming and version labels seemed to be causing issues so i uploaded again with the correct name and version labels - the old model has been taken care of appropriately

luca-martial:
@vkocaman i changed the text to make it longer as you suggested and changed the metrics format to the new one we've been told to use, should be ready to merge

PR #9776, opened 2022-06-24
2022-06-24-xlnet_large_token_classifier_conll03_pipeline_en_3_0

jsl-models:


josejuanmartinez:
@ahmedlone127 is this name correct ? I thinks e had a convention like architecture_ner_ ... Instead of "token classifier"

ahmedlone127:
image



This is the list I got from Maziyar .

PR #9798, opened 2022-06-26
2022-06-26-icd10cm_snomed_mapper_en_3_0

jsl-models:


galiph (approved):

vkocaman:
@galiph @Damla-Gurbaz are you sure chunkMapper accepts resolver type ?
image

galiph:
> @galiph @Damla-Gurbaz are you sure chunkMapper accepts resolver type ? image

Only, in LightPipeline. When we use transform, it fails. @vkocaman

PR #9807, opened 2022-06-26
2022-06-26-abbreviation_mapper_en_3_0

jsl-models:


vkocaman:
@galiph  which dataset is being used here ? abbreviation mapping cannot be static as the meaning can change given the context.. that's why we have `sbiobertresolve_clinical_abbreviation_acronym` . am I missing something here ?

galiph (approved):

galiph:
> 

We used the dataset in our dataset library. We are updating the pretrained model after the changes in ChunkMapper. 
https://nlp.johnsnowlabs.com/2022/05/11/abbreviation_mapper_en_3_0.html
https://www.johnsnowlabs.com/marketplace/list-of-abbreviations-and-acronyms-for-medical-regulatory-activities/

PR #9872, opened 2022-06-29
2022-06-29-dependency_parse_en_3_0

jsl-models:


maziyarpanahi:
@danilojsl this PR should be closed right?

danilojsl:
Hi @maziyarpanahi, no for version 4.0.0. It requires retraining this pipeline model I think it is because of the new scala version used in 4.0.0.

maziyarpanahi:
@danilojsl I think we are good now without this model? https://github.com/maziyarpanahi/spark-nlp-py-starter/actions/runs/2605675326

danilojsl:
@maziyarpanahi it looks like that, but if the tested code has a `pretrained()` call it's because the pipeline model for 3.4.4 and 4.00 is already on S3

maziyarpanahi:
@danilojsl I forgot this is a pipeline. So could you please have a look at the conflict, it seems something with the same name was uploaded already and merged

danilojsl:
Sure, I resolved the conflict

PR #9901, opened 2022-07-05
FEATURE NMH-103: Improvement benchmarking parser [skip-test]

KshitizGIT:
Jira Ticket [NMH-103](https://johnsnowlabs.atlassian.net/browse/NMH-103)

### DIFF
--- a/docs/_plugins/search_index.rb
+++ b/docs/_plugins/search_index.rb
@@ -88,6 +88,12 @@ class Extractor
     nil
   end
 
+  def print_error(message)
+    if ENV["DEBUG"]
+      print(message + "\n")
+    end
+  end
+
   def predicted_entities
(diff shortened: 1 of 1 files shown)


docs/_plugins/search_index.rb
```diff
@@ -141,7 +141,13 @@ def benchmarking_results(post_url)
               row_data = {}
               values = line.split
               values.each_with_index do |value, j|
```
pabla, line 143: Could we also check that the number of cells in the line matches the number of columns?

KshitizGIT: Done

pabla (approved):

maziyarpanahi (approved):

PR #10669, opened 2022-07-21
2022-07-21-bert_pos_bert_base_swedish_cased_pos_sv_3_0

jsl-models:


gadde5300:
predictedEntities should be removed.

PR #10670, opened 2022-07-21
2022-07-21-bert_pos_bert_english_uncased_finetuned_chunk_en_3_0

jsl-models:


gadde5300:
predicted entities should be removed.

PR #10680, opened 2022-07-24
2022-07-24-clinical_deidentification_glove_en_3_0

jsl-models:


vkocaman:
@luca-martial lets call this _slim, not _glove

luca-martial:
ok, renaming and reuploading

PR #10719, opened 2022-07-25
2022-07-25-assertion_oncology_treatment_binary_wip_en_3_0

jsl-models:


vkocaman:
@mauro-nievoff @Cabir40  why do we have nlu mentioned over there ?

PR #10726, opened 2022-07-26
2022-07-26-ner_medication_pipeline_en_3_0

jsl-models:


vkocaman:
@galiph can we add a chunk with dosage in the example?

galiph:
@vkocaman I added dosage to some drugs. Can you check it?

PR #10734, opened 2022-07-27
2022-07-27-robert_sequence_classifier_self_reported_stress_tweet_en_3_0

jsl-models:


muhammetsnts (approved):

vkocaman:
@muhammetsnts  @hsaglamlar  We cannot have the classes as 0 and 1. They should be human readable.. Result column should also be the second column, not the first..

muhammetsnts:
OK, this model should be re-trained. So closed.

PR #10736, opened 2022-07-27
2022-07-27-bert_sequence_classifier_self_reported_vaccine_status_tweet_en_3_0

jsl-models:


vkocaman:
why do we have the result column as the first column ?

PR #12621, opened 2022-08-23
2022-08-23-drug_ade_mapper_en

jsl-models:


Ahmetemintek:
drug-ade mapper from FDA data.

vkocaman:
@Ahmetemintek good one !! will we have the rxnorm_ade_mapper as well ?

galiph:
> @Ahmetemintek good one !! will we have the rxnorm_ade_mapper as well ?

Yes, we will 👍

PR #12747, opened 2022-09-15
2022-09-15-assertion_jsl_augmented_en

jsl-models:


Ahmetemintek:
an augmented version of the assertion_jsl model, also confidence scores were included during training.

muhammetsnts (approved):

PR #12798, opened 2022-09-24
Feature/detect missing class in jvm

C-K-Loan:
## Description
Added function for detection of missing ClassPaths in JVM and printing of warning statements
Integrated calls to this function just before every call to `_new_java_obj()` in the spark NLP python API. 
Whenever an Annotator/Pipeline is constructed, the `check_obj_exist` function will check if the classpath is actually defined in the JVM.  If not, a warning message is printed. 

We could optionally throw an exception when printing the warning, but py4j should throw an exception anyways right after the warning. 

## Motivation and Context
Reduce the amount of problems people have when working with Spark NLP and other John Snow Labs software which depends on it 

## How Has This Been Tested?
Tested in this notebook https://colab.research.google.com/drive/1My4vgLJdd6DUMC-xaLGnzMnZMhcT3lDc?usp=sharing
The jsl lib has easy utilities to include/exclude jars and test this behavior.

image

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] All new and existing tests passed. --> did not run deep tests, but this should not be breaking anything


maziyarpanahi:
@C-K-Loan I have updated the error message. We have to make sure users understand what's missing really, and also a link to the GitHub usage for reference.

This, however, seems to be cosmetic and not trivial. You don't mind if we put this on hold and release it in 4.2.1? (it's a bit last minute and it's a major release where we have done most of the tests)

C-K-Loan:
@maziyarpanahi  sure there is n rush to merge this, we can add it in the next release. 

I agree those edits are pretty deep in the source and could easily break stuff.
I mad really sure it should not be able to break anything but sure, let's keep it for the next release.


And sure makes sense, Ideally, the Error Message would tell you exactly which jar is missing and where to get it.
For this, we need to be able to infer from just looking at the classpath, if the annotator belongs to license or open source.  

Let's see how I can improve this until next release

maziyarpanahi:
> @maziyarpanahi sure there is n rush to merge this, we can add it in the next release.
> 
> I agree those edits are pretty deep in the source and could easily break stuff. I mad really sure it should not be able to break anything but sure, let's keep it for the next release.
> 
> And sure makes sense, Ideally, the Error Message would tell you exactly which jar is missing and where to get it. For this, we need to be able to infer from just looking at the classpath, if the annotator belongs to license or open source.
> 
> Let's see how I can improve this until next release

Perfect! I think the best thing is to have a simple discussion (How To) like we did for Windows and have the link in the error. We can explain all the possible scenarios (SparkSession doesn't have the JAR, SparkSession already existed so it didn't take the spark.jars, etc.). 

I will prepare this for the next release.

python/sparknlp/internal/py4j_utils.py
```diff
@@ -0,0 +1,32 @@
+# Coloring the print
+FAIL = '\033[91m'
+ENDC = '\033[0m'
```
danilojsl, line 3: The name of these variables is a little confusing. To make it more readable, we could rename those e.g.
```
FAIL to FAIL_COLOR
ENDC to END_COLOR
```

python/sparknlp/internal/extended_java_wrapper.py
```diff
@@ -33,6 +34,8 @@ def apply(self):
         return self._java_obj
 
     def new_java_obj(self, java_class, *args):
+        # check_obj_exist(java_class) TODO is checking this here efficient?
+        # if this is just used for primitive data types, we might want to ommit check here?
```
danilojsl, line 38: It's not only for primitive types. If we include `check_obj_exist` validation, it will do for all classes defined [here](https://github.com/JohnSnowLabs/spark-nlp/blob/master/python/sparknlp/internal/__init__.py).
In addition, it's used by python classes that read `CoNLL` and `POS` formats (conll.py, conllu.py, pos.py and pub_tator.py)
Not sure about the latency impact. We will need to do performance tests.

python/sparknlp/internal/py4j_utils.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def check_obj_exist(java_class):
```
danilojsl, line 6: If `java_class` argument gets a datatype different than a string, the function will fail.
We could add a data type validation to make sure `java_class` is a string.

python/sparknlp/internal/py4j_utils.py
```diff
@@ -0,0 +1,32 @@
...
+    # only call this after JVM is turned on
+    from pyspark import SparkContext
+    from pyspark.ml.util import _jvm
+    from py4j.java_gateway import UserHelpAutoCompletion
+    sc = SparkContext._active_spark_context
+    java_obj = _jvm()
+    for name in java_class.split("."):
```
danilojsl, line 13: It seems we are making worthless iterations on this loop. Shouldn't be enough just getting `java_obj` of the last element of `java_class.split(".")` ?
i.e. `name = java_class.split(".")[-1]`

danilojsl (changes requested):

maziyarpanahi:
Thanks @danilojsl for the review. @C-K-Loan The ratio of risk/test to what we gain (a simple error that suggests what's missing) is a bit high. Would it be possible to look for `JavaPackage` error at your end and display relavnt message there instead of here?

python/sparknlp/internal/py4j_utils.py
C-K-Loan: @danilojsl 
I took this iteration from the PySpark source code, it seems that the way they resolve a classpath in Python and create a object in the JVM
[here]https://github.com/apache/spark/blob/4201a5924b5bd12d2c89f31b6049272748c58728/python/pyspark/ml/wrapper.py#L83  


This looks redundant, but we need to check for all the sub-class-path strings if they exsist.
I.e. if the classpath is `A.B.C.D` we must check 
- `A` 
- `A.B`
- `A.B.C`
- `A.B.C.D`

In python we can have `A.B.C.D` defined, but the other py-class-paths may not be defined and will cause JVM not to find class in PVM

python/sparknlp/internal/extended_java_wrapper.py
C-K-Loan: Thats why I suggest to ommit it  there but keep it for the other places where we just do it once for Anno Classes/  High Level Classes which are usually just created once

C-K-Loan:
@maziyarpanahi @danilojsl some alternative thoughts I have, which should be a bit more efficient and will cut not that deep into the source is the following : 

Alternative 1: 
- Get list of references to every Anno-Py-Class
- Update their `__init__` method dynamically with a try/catch error. then `JavaObjectNotCallable` is thrown during `__init__` we can catch it and then do the `check_obj_exist()` I defined for the anno-class.

Alternative 2: 
Another alternative is introducing a function wrapper, something like `@catch_java_obj_not_callabe()`  which will do the try/catch wrapping. This would make our code more verbose though if we have to add this statement to all classes by hand. 



Giving Alternative 1 a shot

maziyarpanahi:
> @maziyarpanahi @danilojsl some alternative thoughts I have, which should be a bit more efficient and will cut not that deep into the source is the following :
> 
> Alternative 1:
> 
> * Get list of references to every Anno-Py-Class
> * Update their `__init__` method dynamically with a try/catch error. then `JavaObjectNotCallable` is thrown during `__init__` we can catch it and then do the `check_obj_exist()` I defined for the anno-class. Essentially Monkey Patching them
> 
> Alternative 2: Another alternative is introducing a function wrapper, something like `@catch_java_obj_not_callabe()` which will do the try/catch wrapping. This would make our code more verbose though if we have to add this statement to all classes by hand.
> 
> Giving Alternative 1 a shot

But both alternatives require changes in the code base. As long as we introduce something new in the code base the risk of any latency or future issues outweighs the benefit that is pretty much telling the user `you might miss including spark-nlp artifacts in the SparkSession`. (knowing the users will come and ask nonetheless)

Is there an alternative to catch the error outside `spark-nlp`, do a pattern match for `JavaPackage` and display extra info there?

C-K-Loan:
Checking during `__init__()` or just catching an exception around `__init__()` will have barely any effect on latency and will not influence any benchmarks, since we don't benchmark the construction of annotators but the transforming with already constructed once. 

Wrapping `__init__()` with a try/except can barely break anything, especially if we just print a message and then raise the exception to the next level.  We are not handling this error and this would just be a handler that does nothing besides printing and re-raising. 
The user's python process would crash anyways since an exception is thrown unless users are actually catching that error, which is why we should still raise the error but at least do the logging.


The error message can be improved a lot and even point to a dedicated documentation page, which explains the error and solves it for them. It could save a lot of time for us


The only way to catch this error in a generic way, such that it affects all users is to add a try-catch around the annotator construction. 

Theoretically, the monkey patch can be performed only in `johnsnowlabs` library side and then there are no changes to source code, but only all users of `johnsnowlabs` libs get this, but I am not sure we would like to split this into the other libs and not have this tested within our default CI pipe. 

My idea was to provide this feature inside of `spark-nlp` so everyone benefits from it.
I glad to improve this PR in such a way that it only affects Annotator Construction calls and make sure no benchmarks are degraded and all tests are passing

PR #12865, opened 2022-09-30
2022-09-30-icd10_icd9_mapper_en

jsl-models:


Ahmetemintek:
This mapper takes icd10 codes and maps it with the corresponding icd9 code.

PR #12875, opened 2022-10-02
FIX: ContextSpellCheckerModel updateVocabClass Nullpointer

C-K-Loan:
## Description

ContextSpellCheckerModel throws nullPointer exception when updateVocabClass with an array and `append=true`. 
For some reason the `vocab` attribute of the specialTransducers classes is not properly initialized.

## How Has This Been Tested?
Tested in Scala, added new test with re-produces the error in Spark-NLP 4.2.0 and added fix with a simple null check.

### DIFF
--- a/src/main/scala/com/johnsnowlabs/nlp/annotators/spell/context/ContextSpellCheckerModel.scala
+++ b/src/main/scala/com/johnsnowlabs/nlp/annotators/spell/context/ContextSpellCheckerModel.scala
@@ -30,6 +30,7 @@ import org.slf4j.LoggerFactory
 
 import java.util
 import scala.collection.JavaConverters._
+import scala.collection.mutable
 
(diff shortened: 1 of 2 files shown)


maziyarpanahi (approved):

albertoandreottiATgmail:
@C-K-Loan can you explain if the null value appears at creation time? or after persisting the model and loading it again? Please give more detail, it seems you're masking the symptom instead of curing the disease.

C-K-Loan:
@albertoandreottiATgmail  
Its occurs when using a pre-trained model and trying to set the parameter.
When running on latest Spark NLP version, you can use this notebook to reproduce https://github.com/JohnSnowLabs/spark-nlp-workshop/blob/master/tutorials/Certification_Trainings/Public/7.Context_Spell_Checker.ipynb

PR #12953, opened 2022-10-19
2022-10-19-ner_covid_trials_en

jsl-models:


Ahmetemintek:
Model was re-trained with the Vaccine_Name label

Issue #12981, opened 2022-10-24
Multiple Documents Do Not Get Chunked Properly


dataanalyst4lyfe:
## Description

If I load in multiple documents, and attempt to chunk them, I get really strange results.
Here is a [diagram of pipeline](https://i.imgur.com/tb9WRQQ.png) that I have created.
Here is the code that I am attempting to run that gives the issue:
```
data = spark.createDataFrame([["Neopets - The gaming website for all your needs.", "Peter Pipers employees are picking pecks of pickled peppers"]]).toDF("title", "description")
data.show()

# Create pipeline

# Document Assembler
# Title
title_document_assembler = DocumentAssembler()
title_document_assembler.setInputCol('title')
title_document_assembler.setOutputCol('title_document')
# Description
description_document_assembler = DocumentAssembler()
description_document_assembler.setInputCol('description')
description_document_assembler.setOutputCol('description_document')

# Sentence Detector - Combining both [Title] & [Description]
sentence_detector = SentenceDetector()
sentence_detector.setInputCols(['title_document', 'description_document'])
sentence_detector.setOutputCol('title_description_sentences')

# Tokenizer
tokenizer = Tokenizer()
tokenizer.setInputCols(['title_description_sentences'])
tokenizer.setSplitChars(['|'])
tokenizer.setOutputCol('title_description_tokens')

# Normalizer
if False:
    normalizer = Normalizer()
    normalizer.setInputCols(['title_description_tokens'])
    normalizer.setCleanupPatterns(['[^\w\d\s]'])
    normalizer.setOutputCol('title_description_normalized')

# POS Tagger
pos_tagger = PerceptronModel.pretrained("pos_anc", "en")

pos_tagger.setInputCols(['title_description_tokens', 'title_description_sentences'])
pos_tagger.setOutputCol('title_description_pos')    
    
# Chunker
chunker = Chunker()
chunker.setInputCols(['title_description_sentences', 'title_description_pos'])
chunker.setRegexParsers([
    '<JJ>*<NNP>+',
])
chunker.setOutputCol('title_description_chunks')

# Finisher
finisher = Finisher()
finisher.setInputCols(['title_document',
                       'description_document',
                       'title_description_sentences',
                       'title_description_tokens',
                       'title_description_pos', 
                       'title_description_chunks'])
finisher.setCleanAnnotations(False)

recursive_pipeline_stages = [
    title_document_assembler,
    description_document_assembler,
    sentence_detector,
[...]

PR #12997, opened 2022-10-27
Adding Control for Annotators with One Column

danilojsl:
## Description

Adds inputAnnotatorTypes in Python to identify annotators that expect only one column. This allows handling more than one-column inputs in annotators that only process one column by taking only the first one and ignoring the other columns.

## Motivation and Context

Check issue #12981 

- Local Unit Tests
- Google Colab

## Types of changes

## Checklist:

### DIFF
--- a/python/sparknlp/annotator/audio/wav2vec2_for_ctc.py
+++ b/python/sparknlp/annotator/audio/wav2vec2_for_ctc.py
@@ -15,6 +15,7 @@
 """Contains classes concerning Wav2Vec2ForCTC."""
 
 from sparknlp.common import *
+from sparknlp.common.annotator_type import AnnotatorType
 
 
 class Wav2Vec2ForCTC(AnnotatorModel,
@@ -85,6 +86,8 @@ class Wav2Vec2ForCTC(AnnotatorModel,
     +------------------------------------------------------------------------------------------+
     """
     name = "Wav2Vec2ForCTC"
+
+    inputAnnotatorTypes = [AnnotatorType.AUDIO]
 
     configProtoBytes = Param(Params._dummy(),
                              "configProtoBytes",
--- a/python/sparknlp/annotator/chunker.py
+++ b/python/sparknlp/annotator/chunker.py
@@ -13,6 +13,7 @@
 #  limitations under the License.
 """Contains classes for the Chunker."""
 from sparknlp.common import *
+from sparknlp.common.annotator_type import AnnotatorType
 
 
 class Chunker(AnnotatorModel):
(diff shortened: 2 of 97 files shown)


dataanalyst4lyfe:
Wait, if I am understanding you correctly:

> by taking only the first one and ignoring the other columns.

That means if I did something like this:
```
annotator.setInputCols(["col_1", "col_2"])
```

When that is executed, only results for `col_2` will be returned.
^ If that is the case, I think that is a very bad idea. 

*My Proposed Solution*
Throw an exception if the user calls `setInputCols()` <- This is for col*s*. The exception message should tell the user to use the single column version (`setInputCol()`).

*Why?*
I think this implementation would reduce the risk of "silent failures", and makes it explicit what should be done. The user will know that they can only pass in 1 column at-a-time, and they won't risk trying to have to debug and wonder why only 1 column is being processed.

@danilojsl - Let me know what you think.

maziyarpanahi:
> **Understanding the Current Solution** Wait, if I am understanding you correctly:
> 
> > by taking only the first one and ignoring the other columns.
> 
> That means if I did something like this:
> 
> ```
> annotator.setInputCols(["col_1", "col_2"])
> ```
> 
> When that is executed, only results for `col_1` will be returned. ^ If that is the case, I think that is a very bad idea.
> 
> **My Proposed Solution** Throw an exception if the user calls `setInputCols()` <- This is for col**s**. The exception message should tell the user to use the single column version (`setInputCol()`).
> 
> **Why?** I think this implementation would reduce the risk of "silent failures", and makes it explicit what should be done. The user will know that they can only pass in 1 column at-a-time, and they won't risk trying to have to debug and wonder why only 1 column is being processed.
> 
> @danilojsl - Let me know what you think.

Hi @dataanalyst4lyfe 
Yes, your assumption was correct, it will silently skip the other inputCols. We actually discussed this to be in sync with Scala/Java behavior which is a hard exception with an error that says this annotator only accepts 1 inputCol. We were going to implement the exception for the reason you explained perfectly, thanks for reminding me.

danilojsl:
@maziyarpanahi, I made the changes to raise an exception when the number of columns is different than expected. So, now the behavior will be synced with Scala

maziyarpanahi:
> @maziyarpanahi, I made the changes to raise an exception when the number of columns is different than expected. So, now the behavior will be synced with Scala

Thanks @danilojsl the new `raise TypeError` will make it exactly like Scala so we won't have use cases where we are not sure what goes in and out of the pipeline.

maziyarpanahi (approved):

PR #13004, opened 2022-10-29
2022-10-29-icd10cm_mapper_en

jsl-models:


Ahmetemintek:
Takes text(ner_chunk) and returns icd10cm code.

PR #13035, opened 2022-11-04
SPARKNLP 643 detecting spark version in a safer way

maziyarpanahi:
Some environments will have an alphanumeric version of Spark. This PR will make a safer comparison when it comes to whether or not we are in Spark `3.3.0` environment (needed for some operations).

We could use 

```python
cleaned_version = re.findall(r'(?:(\d+\.(?:\d+\.)*\d+))', self.sc.version)
```

or `LooseVersion` from `distutils.version`

```python
LooseVersion(self.spark_version()) >= LooseVersion("3.3.0")
```

### DIFF
--- a/python/sparknlp/internal/extended_java_wrapper.py
+++ b/python/sparknlp/internal/extended_java_wrapper.py
@@ -16,6 +16,7 @@
 from pyspark import SparkContext
 from pyspark.ml.wrapper import JavaWrapper
 from pyspark.sql import DataFrame
+import re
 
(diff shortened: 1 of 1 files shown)


maziyarpanahi:
@danilojsl I have 2 commits representing both approaches, they both work even in EMR. Please let me know which one you are more comfortable with

danilojsl:
@maziyarpanahi the one with LooseVersion LGFM

PR #13048, opened 2022-11-08
Fix a bug in Vision Transformer annotator that results in NaNs for some models

ahmedlone127:
## Description
A lot of the VIT models were prdicting NaNs because in their labels some were saved as BigInts while others were saved as Strings , I added code for corss checking

## How Has This Been Tested?
Tested it locally on models that were predicting Nan before the change

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorflowViTClassifier.scala
+++ b/src/main/scala/com/johnsnowlabs/ml/tensorflow/TensorflowViTClassifier.scala
@@ -127,9 +127,13 @@ class TensorflowViTClassifier(
         val logits = tag(encoded, activation)
 
         batch.zip(logits).map { case (image, score) =>
-          val label =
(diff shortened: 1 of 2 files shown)


maziyarpanahi:
@ahmedlone127 
Could you please also have a look at the notebook we have that imports ViT model to Spark NLP to be sure we force either Int or String as the Label type? (not leaving it to interpretations)

ahmedlone127:
image
this is the code for saving the labels, we don't force a specific type here, The only way to be sure the only types allowed are string or Int would be to either test all the spark-nlp VIT models after pushing this change or  reimporting everythig after converting the labels in the config to String  or Int

ahmedlone127:
The code I pushed right now helps support both Int and String therfore we should be good.

maziyarpanahi (approved):

PR #13085, opened 2022-11-15
2022-11-15-longformer_legal_embeddings_es

jsl-models:


josejuanmartinez (changes requested):
The name of the model in the description does not match the real name of the model

josejuanmartinez (approved):

PR #13095, opened 2022-11-16
2022-11-16-abbreviation_category_mapper_en

jsl-models:


Ahmetemintek:
Takes abbreviation, returns definition and category.

PR #13113, opened 2022-11-18
2022-11-18-kegg_disease_mapper_en

jsl-models:


Ahmetemintek:
Takes diseases, returns category, description, icd10, icd11, mesh, and brite codes.

PR #13126, opened 2022-11-22
fixed Wrong Example code provided for LemmatizerModel #13125

### LINKED ISSUE #13125: Wrong Example code provided for LemmatizerModel
In the example code (both Python and Scala) the setInputCols wrongly has 2 inputs. It should be only `token`:

- example: https://nlp.johnsnowlabs.com/2022/03/31/lemma_taiga_ru_3_0.html
- https://nlp.johnsnowlabs.com/api/python/reference/autosummary/python/sparknlp/annotator/lemmatizer/index.html#python.sparknlp.annotator.lemmatizer.LemmatizerModel
- https://colab.research.google.com/drive/1q_PS_IeDTVIEVabnbquk3cLXpoJgFwXy?usp=sharing

Reference issue: https://github.com/JohnSnowLabs/spark-nlp/issues/13124

ahmedlone127:
fixes #13125 by changing example provided in modelshub

### DIFF
--- a/docs/_posts/maziyarpanahi/2022-03-31-lemma_alksnis_lt_3_0.md
+++ b/docs/_posts/maziyarpanahi/2022-03-31-lemma_alksnis_lt_3_0.md
@@ -46,7 +46,7 @@ tokenizer = Tokenizer()\
     .setOutputCol("token") 
 
 lemma = LemmatizerModel.pretrained("lemma_alksnis", "lt")\ 
-    .setInputCols(["sentence", "token"])\ 
+.setInputCols(["token"])\
     .setOutputCol("lemma")
     
 pipeline = Pipeline(stages=[document, sentence, tokenizer, lemma])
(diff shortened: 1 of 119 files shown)


maziyarpanahi:
@ahmedlone127 
- we cannot remove SentenceDetectorDL, we are using that inside the `Tokenizer()`
- also, I believe all the models based on `LemmatizerModel` have the wrong number of `.setInputCols`. Everything here: https://nlp.johnsnowlabs.com/models?task=Lemmatization&edition=Spark+NLP&type=model

ahmedlone127:
changing by hand would be time consuimg, would a progmmatic approach be more favorable?

maziyarpanahi:
> changing by hand would be time consuimg, would a progmmatic approach be more favorable?

Definitely! Once you have that script we will extend it to even check all the `inputCols` to be sure we are setting them correctly. 

We can start with Python first, once it was solid we can then fix the Scala (if any) right here int this branch.

ahmedlone127:
Alright, I will work on this tommorow

PR #13243, opened 2022-12-16
2022-12-16-legclf_survival_bert_en

jsl-models:


josejuanmartinez (commented):
Hey @dcecchini ! Can you try to reformat some of the Python Code sections? Ideally we don't include any tabulation, we start with the code directly. This is how I see some of the files:

image

josejuanmartinez (approved):

PR #13283, opened 2022-12-28
SPARKNLP-667: Fix indexing issue for custom pattern

DevinTDHa:
## Description

This PR fixes an issue for the Tokenizer, when a custom pattern is used with lookahead/-behinds that have 0 width matches, indexes would not be calculated correctly. This can lead to further issues down a pipeline.

Other changes:

- resolved some warnings
- refactored tokenizer tests and added new index alignment check

## How Has This Been Tested?

Existing tests are passing (and appended with an additional index check), new tests have been added to cover this scenario

## Types of changes

### DIFF
--- a/src/main/scala/com/johnsnowlabs/nlp/annotators/TokenizerModel.scala
+++ b/src/main/scala/com/johnsnowlabs/nlp/annotators/TokenizerModel.scala
@@ -254,8 +254,10 @@ class TokenizerModel(override val uid: String)
   private val PROTECT_CHAR = "ↈ"
   private val BREAK_CHAR = "ↇ"
 
-  private lazy val BREAK_PATTERN = ("[^(?:" + $(targetPattern) + ")" + PROTECT_CHAR + "]").r
-  private lazy val SPLIT_PATTERN = ("[^" + BREAK_CHAR + "]+").r
+  private lazy val BREAK_PATTERN: Regex =
+    ("[^(?:" + $(targetPattern) + ")" + PROTECT_CHAR + "]").r
+  private lazy val SPLIT_PATTERN: Regex =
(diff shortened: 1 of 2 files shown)


src/main/scala/com/johnsnowlabs/nlp/annotators/TokenizerModel.scala
```diff
@@ -349,21 +351,17 @@ class TokenizerModel(override val uid: String)
...
-                          try {
-                            IndexedToken(
-                              str,
-                              text.start + candidate.start + curPos,
-                              text.start + candidate.start + curPos + str.length - 1)
-                          } finally {
-                            curPos += str.length + 1
-                          })
-                      } finally {
-                        curPos -= 1
```
DevinTDHa, line 366: I don't know why initially there were these `try-finally` blocks without catching any exceptions or handling any resources. I removed them as we would want to be specific with the exceptions (and easier maintenance if a bug would occur here).

maziyarpanahi (approved):

src/main/scala/com/johnsnowlabs/nlp/annotators/TokenizerModel.scala
DevinTDHa: @maziyarpanahi 

It was part of the issue, due to the way the indexes were calculated (always adding and subtracting 1). 

However, I think that having a log line there would be pretty useful. I will add it in a new commit!

DevinTDHa: In case it is needed I openen the PR here:
https://github.com/JohnSnowLabs/spark-nlp/pull/13291

PR #13432, opened 2023-01-30
SPARKNLP-733: Fix loadSavedModel for private buckets

DevinTDHa:
## Description

This PR fixes an issue, where a private bucket S3 would fail to download, when using loadSavedModel.

## How Has This Been Tested?

Ran relevant tests with a public and private S3 bucket. 

## Types of changes

### DIFF
--- a/src/main/scala/com/johnsnowlabs/nlp/pretrained/ResourceDownloader.scala
+++ b/src/main/scala/com/johnsnowlabs/nlp/pretrained/ResourceDownloader.scala
@@ -591,11 +591,19 @@ object ResourceDownloader {
     val (bucketName, keyPrefix) = ResourceHelper.parseS3URI(path)
 
     val (accessKey, secretKey, sessionToken) = ConfigHelper.getHadoopS3Config
-    val privateS3Defined = accessKey != null && secretKey != null && sessionToken != null
+    val region = ConfigLoader.getConfigStringValue(ConfigHelper.awsExternalRegion)
+    val privateS3Defined =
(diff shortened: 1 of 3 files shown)


src/main/scala/com/johnsnowlabs/nlp/pretrained/ResourceDownloader.scala
```diff
@@ -589,9 +590,14 @@ object ResourceDownloader {
 
     val (bucketName, keyPrefix) = ResourceHelper.parseS3URI(path)
 
-    val tmpDirectory = SparkFiles.getRootDirectory()
+    val (accessKey, secretKey, sessionToken) = ConfigHelper.getHadoopS3Config
+    val privateS3Defined = accessKey != null && secretKey != null && sessionToken != null
```
danilojsl, line 594: To be 100% sure all private arguments are defined. I will also check the region has data i.e. check `ConfigHelper.awsExternalRegion` is set

danilojsl (changes requested):

src/main/scala/com/johnsnowlabs/nlp/pretrained/ResourceDownloader.scala
DevinTDHa: I have included a check for region now. Also added a log line so that when config is not set correctly the user will be notified:

https://github.com/JohnSnowLabs/spark-nlp/pull/13432/files#diff-053c39e8ab751a99f5c1bb1da5c3b4bfde8954313ea782311a8bdecb4ca44114R593-R606

PR #13452, opened 2023-02-01
Fix to white space in cache_pretrained

dcecchini:
## Description

When using `ResourceDownloader` in a Windows environment where the user folder contains white space, the downloader raises an error 

```java
java.net.URISyntaxException: Illegal character in path at index 20: file:/C:/Users/User Name/cache_pretrained/  
```

I used the `java.nio.file.Paths` module to propose a fix, but it needs to be further tested.

## Motivation and Context

Crashes the process of all pipelines.

## How Has This Been Tested?

No tests were made.

## Types of changes

## Checklist:


maziyarpanahi:
I am sorry, the space and other characters inside the full path are problematic in many other places. (meaning we can make some changes here, tests, risk whether or not everything would work, but it will come back when you use spark.read or spark ml load, etc.)

The PATH must follow Linux/Unix convention, no space.

PR #13554, opened 2023-02-22
2023-02-22-ndc_drug_brandname_mapper_en

jsl-models:


Ahmetemintek:
A mapper model that takes NDC codes and returns drug brand names

Ahmetemintek:
Entity resolver for Rxnorm from NIH database.

Ahmetemintek:
A mapper model takes drugs and returns Rxorm codes according to NIH database

PR #13703, opened 2023-03-27
FEATURE NMH-166: Add website to elastic search for search service

KshitizGIT:
Part of Jira Ticket [NMH-166](https://johnsnowlabs.atlassian.net/browse/NMH-166)

### DIFF
--- a/docs/_plugins/search_index.rb
+++ b/docs/_plugins/search_index.rb
@@ -9,10 +9,17 @@ require 'nokogiri'
 SEARCH_URL = (ENV["SEARCH_ORIGIN"] || 'https://search.modelshub.johnsnowlabs.com') + '/'
 ELASTICSEARCH_INDEX_NAME = ENV["ELASTICSEARCH_INDEX_NAME"] || 'models'
 
+WEBSITE = ENV["WEBSITE"] || ""
+
 OUTDATED_EDITIONS = ['Spark NLP 2.0', 'Spark NLP 2.1', 'Healthcare NLP 2.0']
+SPARK_NLP_ORIGIN = "https://sparknlp.org"
+JOHNSNOWLABS_ORIGIN = "https://nlp.johnsnowlabs.com"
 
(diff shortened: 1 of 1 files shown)


docs/_plugins/search_index.rb
```diff
@@ -124,6 +131,11 @@ def print_error(message)
     end
   end
 
+  def website
```
pabla, line 134: Can we use `origin` here?

docs/_plugins/search_index.rb
```diff
@@ -9,10 +9,17 @@
...
 
+WEBSITE = ENV["WEBSITE"] || ""
+
 OUTDATED_EDITIONS = ['Spark NLP 2.0', 'Spark NLP 2.1', 'Healthcare NLP 2.0']
+SPARK_NLP_ORIGIN = "https://sparknlp.org"
+JOHNSNOWLABS_ORIGIN = "https://nlp.johnsnowlabs.com"
 
 $remote_editions = Set.new
-
+if WEBSITE.empty?
```
pabla, line 19: It will break a local server.

KshitizGIT: You could pass `WEBSITE` in environment variable. Alternative it using default value, but which should we use default ?

docs/_plugins/search_index.rb
KshitizGIT: Done

docs/_plugins/search_index.rb
pabla: I mean people who edit the website will face this problem.
This variable is required in required for Elasticsearch. We could also check if ELASTICSEARCH_URL is not empty and only in this case stop the execution.

KshitizGIT: Ok. I added the check only when ELASTICSEARCH_URL is available

pabla (approved):

maziyarpanahi (approved):

PR #13741, opened 2023-04-06
FEATURE NMH-168: Hide licensed models from spark-nlp search [skip-test]

KshitizGIT:
Jira Ticket [NMH-168](https://johnsnowlabs.atlassian.net/browse/NMH-168)

Requires this PR to work https://github.com/JohnSnowLabs/models-hub/pull/158

[NMH-168]: https://johnsnowlabs.atlassian.net/browse/NMH-168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/_frontend/webpack.config.js
+++ b/docs/_frontend/webpack.config.js
@@ -111,6 +111,7 @@ module.exports = (env, argv) => {
         SEARCH_ORIGIN:
           process.env.SEARCH_ORIGIN ||
           'https://search.modelshub.johnsnowlabs.com',
+        ORIGIN: process.env.ORIGIN,
       }),
       isProduction &&
         new MiniCssExtractPlugin({
--- a/.github/workflows/create_search_index.yml
+++ b/.github/workflows/create_search_index.yml
@@ -75,6 +75,7 @@ jobs:
         working-directory: docs/_frontend
         env:
           SEARCH_ORIGIN: ${{ secrets.SEARCH_ORIGIN }}
+          ORIGIN: ${{ secrets.ORIGIN }}
         run: |
           yarn
           yarn run build
--- a/docs/_frontend/src/models/App/App.js
+++ b/docs/_frontend/src/models/App/App.js
@@ -116,7 +116,7 @@ const useFilterQuery = () => {
     window.history.pushState(
       nextParams,
       window.title,
-      '/models' + toSearchString(nextParams)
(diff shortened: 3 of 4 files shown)


maziyarpanahi (approved):

pabla (commented):
Let's make it simpler.

We can add a condition to the `toSearchString` method:
```js
  if (ORIGIN) {
    searchParams.append('origin', ORIGIN);
  }
```
And we don't need to modify the App and Sidebar components.

docs/_frontend/webpack.config.js
```diff
@@ -111,6 +111,7 @@ module.exports = (env, argv) => {
         SEARCH_ORIGIN:
           process.env.SEARCH_ORIGIN ||
           'https://search.modelshub.johnsnowlabs.com',
+        ORIGIN: process.env.ORIGIN || 'https://sparknlp.org',
```
pabla, line 114: I think we could remove a default value from here.

KshitizGIT:
> Let's make it simpler.
> 
> We can add a condition to the `toSearchString` method:
> 
> ```js
>   if (ORIGIN) {
>     searchParams.append('origin', ORIGIN);
>   }
> ```
> 
> And we don't need to modify the App and Sidebar components.

If we just modify the `toSearchString` , the origin=https://sparknlp.org will appear in the browser url .

pabla:
> @pabla , If we just modify the `toSearchString` , the origin=https://sparknlp.org will appear in the browser url . I think it would be better to hide origin=https://sparknlp.org from the url fragments in the browser

Yes, you are right, the origin shouldn't be visible in the browser address.
But I would like to keep all things related to origin in one place.
What if we introduce a new parameter to the `toSearchString` method and don't include `ORIGIN` when we generate an URL for the browser address?

```js
window.history.pushState(
  nextParams,
  window.title,
  '/models' + toSearchString(nextParams, false)
);
```

docs/_frontend/webpack.config.js
pabla: I would prefer to keep the frontend code the same as we have in the `johnsnowlabs` repo.

KshitizGIT: Remove default value

pabla (approved):

PR #13818, opened 2023-05-23
[WIP] SPARKNLP-831: Python Dynamic Setters and Getters

DevinTDHa:
## Description

This WIP PR changes, how parameters of annotators will be set on the Python side. The getter and setter functions are dynamically generated. The generated setter functions call the equivalent on on the Scala side, to automatically enable validations implemented there. These functions can still be implemented manually. They will only be generated if they don't exist.

An example of how the new annotators would look like, can be seen in the [`Tokenizer`](https://github.com/JohnSnowLabs/spark-nlp/blob/3b91e580b3560a42e0872ee8d4edde37207c5d6e/python/sparknlp/annotator/token/tokenizer.py) class. We only need to declare the parameters, and the corresponding getters and setters (with doc string) will be accessible. Only functions with special behavior will still need to be implemented.

TODO:

- [ ] Only Tokenizer implemented so far
- [x] Type checks for provided arguments

### How it works

This is done with the `ParamsGettersSetters` mix-in (which already existed). While it already generated getters and setters which didn't exist in the class, it tried to set parameters directly via Spark. This was modified to call the equivalent one on the Scala side instead:

https://github.com/JohnSnowLabs/spark-nlp/blob/3b91e580b3560a42e0872ee8d4edde37207c5d6e/python/sparknlp/internal/params_getters_setters.py#L145-L164

Py4J automatically converts primitive types, when trying to call a java function. However, when trying to pass a list as a parameter, Py4J would convert to an `ArrayList`. This would clash with the signature on the Scala side, as the functions require an `Array`. A new [`converter.py`](https://github.com/JohnSnowLabs/spark-nlp/blob/3b91e580b3560a42e0872ee8d4edde37207c5d6e/python/sparknlp/internal/converters.py) module implements the necessary conversion for it to work as expected.

Note/TODO: We might want to change the existing class [ExtendedJavaWrapper](https://github.com/JohnSnowLabs/spark-nlp/blob/ccce219e7bb1fd7185bdf77fca319ceb819f8936/python/sparknlp/internal/extended_java_wrapper.py#L22) to incorporate this behavior. I'm not sure if these functionalities are used anywhere at the moment.

### Tradeoffs
As the functions are dynamically generated

- documentation for the getters and setters will not be available anymore. We will have to add documentation for users to "just" assume they exist
- Code hints will not be available in IDEs (But are available in interactive environments such as jupyter or colab, see screenshots)

## Motivation and Context

Some setter functions on the Scala side have validity checks for the input. These would have to be equivalently implemented on the Python side, leading to unnecessary work (and error proneness). This change will eliminate this need and speed up annotator implementation on the Python side as well.

## How Has This Been Tested?

- New tests added for the generator class. 
- New tests to verify Scala side checks are happening for the Tokenizer. 
- All tests are Passing

Doc Strings in interactive Environment:
image

Verification from Scala side:
https://github.com/JohnSnowLabs/spark-nlp/blob/ccce219e7bb1fd7185bdf77fca319ceb819f8936/src/main/scala/com/johnsnowlabs/nlp/annotators/Tokenizer.scala#L288-L292
image

## Types of changes

## Checklist:

### DIFF
--- a/python/sparknlp/annotator/token/tokenizer.py
+++ b/python/sparknlp/annotator/token/tokenizer.py
@@ -13,7 +13,6 @@
 #  limitations under the License.
 """Contains classes for the Tokenizer."""
 
-
 from sparknlp.common import *
 
 
@@ -130,27 +129,27 @@ class Tokenizer(AnnotatorApproach):
 
     contextChars = Param(Params._dummy(),
                          "contextChars",
-                         "character list used to separate from token boundaries",
+                         "The character list used to separate from token boundaries",
                          typeConverter=TypeConverters.toListString)
 
     splitPattern = Param(Params._dummy(),
                          "splitPattern",
-                         "character list used to separate from the inside of tokens",
+                         "The character list used to separate from the inside of tokens",
                          typeConverter=TypeConverters.toString)
 
     splitChars = Param(Params._dummy(),
                        "splitChars",
-                       "character list used to separate from the inside of tokens",
+                       "The character list used to separate from the inside of tokens",
                        typeConverter=TypeConverters.toListString)
 
     minLength = Param(Params._dummy(),
                       "minLength",
-                      "Set the minimum allowed length for each token",
+                      "The minimum allowed length for each token",
                       typeConverter=TypeConverters.toInt)
 
     maxLength = Param(Params._dummy(),
                       "maxLength",
-                      "Set the maximum allowed length for each token",
+                      "The maximum allowed length for each token",
                       typeConverter=TypeConverters.toInt)
 
     @keyword_only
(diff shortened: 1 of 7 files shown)


DevinTDHa:
Hi @maziyarpanahi 

> How does the Scaladoc looks like in this approach

There are no changes necessary on the Scala side, so it stays as it is right now.

> Is this compatible (aligned) with the way Apache Spark deals with parameters between Scala/Python?

The getter still retrieves the saved parameter from Spark as usual. The setter just calls the Scala side setter, so as long as that one conforms to Spark and is correct, so will be the Python one!

DevinTDHa:
Ah, yes the PyDoc will have the getters setters missing, except for the ones that are explicitly defined (unless I can figure out how sphinx can generate it dynamically).

So we need to make sure all the parameters are documented.  I think the parameters themselves can also be generated but I'll need to change some settings.

The Tokenizer currently looks like this:
image

All the functions you see are explicitly implemented (because of some extra functionality)

maziyarpanahi:
Thanks for the explanation. So in the actual Apache Spark like Spark ML, they are explicitly defining everything on both sides like we do right now? https://spark.apache.org/docs/latest/api/python/_modules/pyspark/ml/feature.html#Tokenizer.setInputCol

PR #13846, opened 2023-06-06
2023-06-06-ner_vop_anatomy_emb_clinical_medium_en

jsl-models:


maziyarpanahi:
@mauro-nievoff this cannot be both open-source and licensed at the same time :)
image

PR #13959, opened 2023-09-06
Fixing onnx saving path bug

ahmedlone127:
FIxes saving of Onnx models on windows 

## Description
Currently, Onnx models can't be saved on windows because of the temporary path being used. This PR fixes the temporary path by replacing the character which can raise an exception while saving 

## How Has This Been Tested?
Tested locally by saving models and reloading them on windows 

## Types of changes

## Checklist:

- [ ] I have read the [CONTRIBUTING](https://sparknlp.org/contribute.html) page.

### DIFF
--- a/src/main/scala/com/johnsnowlabs/ml/onnx/OnnxWrapper.scala
+++ b/src/main/scala/com/johnsnowlabs/ml/onnx/OnnxWrapper.scala
@@ -79,8 +79,8 @@ object OnnxWrapper {
 
   // TODO: make sure this.synchronized is needed or it's not a bottleneck
   private def withSafeOnnxModelLoader(
-                                       onnxModel: Array[Byte],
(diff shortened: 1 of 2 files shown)


DevinTDHa:
@maziyarpanahi @ahmedlone127 I pushed a fix, which addresses the underlying issue.

The reason why it used to work for TF but not for onnx is the following:

In tf we always had a name for the graph already defined (`saved_model.pb`). We don't have that. Somewhere down the line we passed the _whole_ path to `onnxWrapper.saveToFile(onnxFile)`. The method `saveToFile` then receives this whole path and creates another tmp folder with it, which includes the whole path (pretty deep nesting). 

- In unix this is ok as we have paths that start with `/`. 
- In Windows however it starts with the drive. So we have something like `C:/tmp_uuid/C:/tmp_uuid2` which causes this issue.

maziyarpanahi (approved):

PR #13997, opened 2023-09-20
SPARKNLP-765: VisionEncoderDecoder

DevinTDHa:
## Description

This PR introduces the VisionEncoderDecoder annotator. This annotator takes images and produces captions.

Pretrained model uploaded at https://github.com/JohnSnowLabs/spark-nlp/pull/13999

### Beam Search Fix

This PR also includes a potential bug fix to our implementation of the beam search algorithm. Explanation:

In this line we initialize the beams with logprob 0 or -1e-9=-0.000000001 (equivalent probabilty $\exp(\mathrm{-1e-9}) \approx 1$), depending on its position

https://github.com/JohnSnowLabs/spark-nlp/blob/master/src/main/scala/com/johnsnowlabs/ml/ai/util/Generation/Generate.scala#L200

This differs from the transformers implementation though, where the other logprob is initialized to be -1e+9=-1000000000 (equivalent probabilty $\exp(\mathrm{-1e9}) \approx 0$).

https://github.com/huggingface/transformers/blob/v4.33.1/src/transformers/generation/tf_utils.py#L2272

So basically, in our version we add a tiny amount, which results in the almost exact scores for the initial beams. Implementing this change results in the same results for this model as in the transformers implementation.

@maziyarpanahi, @prabod is taking a look if this will affect the Bart annotator. Results are different but I am not sure how else it is affected.

## How Has This Been Tested?

Local tests and new test passing. Tested it also in a colab notebook.

## Types of changes

## Checklist:


prabod (approved):
This is a bug. Thanks @DevinTDHa for identifying and fixing the bug.

maziyarpanahi (approved):

PR #14044, opened 2023-10-26
2023-10-26-ner_test_shareingproject_en

jsl-models:


maziyarpanahi:
@psajan The sample code mentions `clinical` but it is uploaded as `open-source`. Is this a test model? Is it being uploaded in a right place?

PR #14084, opened 2023-12-07
520-release-candidate

maziyarpanahi:
- ONNX support for T5
- use_cache support for newly exported T5 in TF (to make performance close to ONNX version)
- ONNX support for Marian
- new notebook fro structure streaming
- fix a bug for Spark NLP version in Scala


danilojsl (approved):
Tests in Colab, EMR, and Databricks passed ✅

PR #14151, opened 2024-01-27
812 implement de berta for zero shot classification annotator

ahmedlone127:
adding DeBertaForZeroShotClassification ( note needs to be reviewed to see why predictions( Very different from huggingface predictions and largely incorrect )  are not being made correctly  , same as BartForZeroShotClassifications


maziyarpanahi:
@DevinTDHa Could you please have a look and see why the predictions are different from HuggingFace?

DevinTDHa:
Hi @maziyarpanahi @ahmedlone127, I encountered this issue before and propose a solution:

### Underlying Cause

I believe the issue could be part of the following. Note that this might be specific to implementations that I looked at, namely RoBertaForQA and MPNetForQA. It might differ for different models.

1. **Attention mask tensors are wrongly encoded**. They should be an array of ones, but instead we are creating it based on the token ids (token 0 could be a valid ID and not a special token, depending on the model)  https://github.com/JohnSnowLabs/spark-nlp/blob/67e67e1eec9c84c59530f7a43afebb654412cfbd/src/main/scala/com/johnsnowlabs/ml/ai/DeBertaClassification.scala#L410
2. **Logits are not pre-processed before calculating probabilities with softmax**. In the original implementations, special tokens and question tokens are not included in the start/end probabilities. This is done by setting the log-probability for these tokens to -10000 which is 0 probability. https://github.com/JohnSnowLabs/spark-nlp/blob/67e67e1eec9c84c59530f7a43afebb654412cfbd/src/main/scala/com/johnsnowlabs/ml/ai/DeBertaClassification.scala#L330-L346
3. **Final Score is not correctly calculated**. it should be the likelihood (`startScore * endScore`) instead of the mean. https://github.com/JohnSnowLabs/spark-nlp/blob/060cf6dd88920d1d73c6923ef684986f2405daf6/src/main/scala/com/johnsnowlabs/ml/ai/XXXForClassification.scala#L303

### Fix

I fixed these issues for RobertaForQA and the upcoming MPNetForQA by overriding the base functions in the specific implementation file (RobertaClassification.scala and MPNetForClassification.scala). 

They are included in the respective PR: https://github.com/JohnSnowLabs/spark-nlp/pull/14147

I chose not to fix the `XXXForClassification.scala` file, as I don't know if the other annotators are also affected that use it. It could be model specific.

So going forward, I'd suggest looking at my implementation and checking if the scores are better than before. The implementation is here: https://github.com/JohnSnowLabs/spark-nlp/blob/a2d08d6b7145f2bc6cc211fc7a8605cbe192d519/src/main/scala/com/johnsnowlabs/ml/ai/MPNetClassification.scala#L446-L450

Just as a note, the tolerance for the scores of my implementation are not that low (0.01 float tolerance). Close enough I suppose but I believe there might be some numerical issues for the implementation in general.

maziyarpanahi:
- Thanks @DevinTDHa appreciate sharing your experience
- @ahmedlone127 could you please apply the changes and re-test?

ahmedlone127:
Thanks for the detailed explanation @DevinTDHa ! 

I will implement these steps and get back to you @maziyarpanahi

ahmedlone127:
Hello @maziyarpanahi , I took another look at the source code and found that by changing the offset value for the tokenizations (from 1 to 0 ) the predictions from the model became a lot more accurate. 
image

maziyarpanahi (approved):

PR #14201, opened 2024-03-11
Fix missing compatible editions [skip-test]

KshitizGIT:
Some compatible_editions are missing from models.json.
For examples:
https://sparknlp.org/models.json, the `zero_shot_ner_roberta` has only these compatible_editions ["Spark NLP 4.3","Spark NLP"]}. Higher spark nlp versions are missing

### DIFF
--- a/docs/_plugins/search_index.rb
+++ b/docs/_plugins/search_index.rb
@@ -64,7 +64,7 @@ def compatible_editions(editions, model_editions, edition)
     selection_lambda = lambda {|_| false }
   end
 
-  editions = editions.select &selection_lambda
+  selected_editions = editions.select &selection_lambda
   model_editions = model_editions.select &selection_lambda
 
   curr_index = model_editions.index(edition)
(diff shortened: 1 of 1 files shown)


docs/_plugins/search_index.rb
```diff
@@ -64,15 +64,15 @@ def to_product_name(edition)
     selection_lambda = lambda {|_| false }
   end
 
-  editions = editions.select &selection_lambda
+  selected_editions = editions.select &selection_lambda
```
pabla, line 67: Hmm.. I see only renamed variables. How does it solve the issue?

KshitizGIT: The variable `editions` is also in the global scope. Since, we are assigning editions, I suspect that's what causing the issue. 

I ran the entire workflow with a fork of Spark NLP. I was not able to replicate the issue for the `zero_shot_ner_roberta`.

pabla: I don't think it will help because we don't modify the variable value here.
Maybe it's an issue with incremental build?

pabla (approved):

PR #14689, opened 2025-11-07
Adding perfered engine to pretrained and minor fixes

ahmedlone127:
This Pr introduces two new mechanisms for pretrained models:

1. Pretrained models will be filtered by annotator class first, before downloading (in case there are duplicate names for different annotators)
2. a new parameter for .pretrained() called `engine` it will allow users to specify which enigne they would perfer to use and if a model is avaialable in multiple engines it will automatically pick the engine specified otherwise fall back to another 
engine. If no engine is specified the following priority order is used :

Onnx -> Tensorflow -> Openvino -> Unk 

## How Has This Been Tested?
Test locally via unit tests in python and scala
Tested on colab


danilojsl:
@ahmedlone127 Could you please add a demo notebook to showcase the use of the new `engine` parameter?

PR #14695, opened 2025-11-09
fix tags

AbdullahMubeenAnwar:
Ensured that each Markdown file’s engine tag matches the value specified in its engine field.
Removed any extra engine tags from the tags section.

Limitations
Most model cards (60k+) do not have an engine field defined in their YAML front matter.


DevinTDHa:
Related to https://github.com/JohnSnowLabs/spark-nlp/pull/14689

DevinTDHa (approved):

PR #14756, opened 2026-04-14
Update llama_cpp_in_Spark_NLP_LLMEntityExtractor.ipynb

AbdullahMubeenAnwar:


danilojsl:
@AbdullahMubeenAnwar it seems something is missing on Python side. It's working on Scala but not in Python.

PR #14762, opened 2026-05-04
updated GetFewShotexamples to handle multiple cases in MedicalLLMEnti…

ahmedlone127:
There is an issue when pasing exmaple sentences in colab to the annotator since the scala doesn't handle all type of inputs , this PR imporves upon the ability to handle different input cases for GetFewShotExamples


danilojsl:
@ahmedlone127 Could you add more details to the description of this PR? We need it to include on the release notes.

PR #14764, opened 2026-05-05
Add LateChunkEmbeddings annotator

AbdullahMubeenAnwar:
## Description
Implements a new `LateChunkEmbeddings` annotator based on the Late Chunking technique described in [Jin et al. (2024), arXiv:2409.04701](https://arxiv.org/abs/2409.04701).

Unlike `ChunkEmbeddings`, which embeds each chunk in isolation, `LateChunkEmbeddings` expects the upstream token-embedding stage (e.g. `LongformerEmbeddings`, `ModernBertEmbeddings`) to have already processed the **full document** in a single forward pass. This annotator then locates the tokens within each chunk's character span and mean-pools them into a single `SENTENCE_EMBEDDINGS` vector, so every chunk embedding is informed by the complete document context rather than being isolated.

## Motivation and Context
Chunking before embedding (the naive approach) breaks contextual coherence, a chunk that refers back to an earlier entity loses that context entirely. Late Chunking fixes this by running the transformer over the full document first, then pooling token vectors by chunk span. This produces significantly better chunk embeddings for RAG and retrieval use cases with no changes to the underlying model.

## How Has This Been Tested?
local scala + python test and assembled JAR tested on google colab

## Types of changes

## Checklist:

### DIFF
--- a/docs/en/annotator_entries/LateChunkEmbeddings.md
+++ b/docs/en/annotator_entries/LateChunkEmbeddings.md
@@ -1,0 +1,193 @@
+{%- capture title -%}
+LateChunkEmbeddings
+{%- endcapture -%}
+
+{%- capture description -%}
+Produces contextual chunk-level embeddings using the **Late Chunking** technique described in
+[Jin et al. (2024)](https://arxiv.org/abs/2409.04701).
+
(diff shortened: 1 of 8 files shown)


danilojsl:
@AbdullahMubeenAnwar Please elaborate a little more on the description and motivation for this new feature so it can be added to the release notes.

Issue #14784, opened 2026-06-25
ONNX Runtime (CPU) causes memory growth on macOS


nogifeet:
### Is there an existing issue for this?

- [x] I have searched the existing issues and did not find a match.

### What are you working on?

Development of the Spark-OCR library; running test cases locally.

### Current Behavior

Bumping onnxruntime from 1.19.2 to 1.24.3 causes a memory issue with ONNX Runtime inference on macOS. Define a pipeline fit via transform() and trigger spark action , memory usage spikes to 80GB+ and eventually crashes the Spark session.
This behavior is observed only in 1.24.3; 1.19.2 works fine.

<img width="559" height="662" alt="Image" src="https://github.com/user-attachments/assets/9c8ece11-2473-44d9-b6bc-5c168e02cbdd" />

### Expected Behavior

Switching to 1.19.2: the pipeline works perfectly fine with minimal memory growth, and results are successfully collected to the driver.

### Steps To Reproduce

Onnx Model -> "image_text_detector_mem_opt"

    binary_to_image = BinaryToImage() \
        .setInputCol("content") \
        .setOutputCol("image") \
        .setImageType(ImageType.TYPE_3BYTE_BGR)
    
    text_detector = ImageTextDetectorCraft \
        .pretrained("image_text_detector_mem_opt", "en", "clinical/ocr") \
        .setInputCol("image") \
        .setOutputCol("text_regions") \
        .setTextThreshold(0.2) \
        .setSizeThreshold(2) \
        .setScoreThreshold(0.1)
    
    pipeline = PipelineModel(stages=[
        binary_to_image,
        text_detector
    ])
    
    df = spark.read.format("binaryFile").load("./sample.png")
    result = pipeline.transform(df).collect() <--- Should Spike the memory

### Spark NLP version and Apache Spark

Spark-OCR Version --> 6.4.2
Spark-NLP Version --> 6.4.1-rc1/6.4.2
Spark --> 3.4.4

### Type of Spark Application

Scala Application

### Java Version

Zulu 8.54.0.21-CA-macos-aarch64

### Java Home Directory

/Users/macos_home/.sdkman/candidates/java/current

### Setup and installation

sbt

### Operating System and Version

MacBook Pro; Apple M4; 16 GB; Silicon

### Additional Information

Behaviour is same for Scala/Python Application

AbdullahMubeenAnwar:
[...]

PR #14791, opened 2026-07-03
 [SPARKNLP-1400] Pin ONNX Runtime to 1.23.0 to fix memory regression …

AbdullahMubeenAnwar:


danilojsl:
This change solve this issue https://github.com/JohnSnowLabs/spark-nlp/issues/14784

PR #14792, opened 2026-07-08
Clean up ONNX temporary files

mehmetbutgul:
## Description

This PR improves temporary file and resource management during model loading, especially on ONNX models and archive extraction.

### Changes

* Added cleanup of temporary directories after loading ONNX models.
* Added cleanup of temporary directories after loading SentencePiece models.
* Marked temporary ONNX extraction directories for deletion on JVM exit when immediate deletion is not possible.
* Updated ZIP extraction logic to ensure `ZipFile` is always closed using a `try/finally` block.
* Extracted temporary directory deletion into a reusable helper method.

These changes are focused on improving resource cleanup and preventing temporary file accumulation without changing existing functionality.

---

## Motivation and Context

Model loading and archive extraction create temporary files and directories that were not always cleaned up properly.
**Previously, loading a 1 GB ONNX model would leave approximately 2 GB of temporary files on disk.**

This PR improves resource management by:

* preventing temporary directory accumulation,
* ensuring extracted temporary resources are deleted when they are no longer needed,
* guaranteeing that `ZipFile` handles are properly released even if an exception occurs during extraction.

No functional behavior of model loading has been changed.

---

## How Has This Been Tested?

Tested by loading different onnx models by using local machine - Windows

Verified that:

* ONNX models load successfully.
* SentencePiece models load successfully.
* ZIP archives are extracted correctly.
* Temporary directories are removed after use (or scheduled for deletion on JVM exit when required).
* No regressions were observed in the model loading workflow.

## Types of changes

## Checklist:


mehmetbutgul:
It would be good to verify that executors are still running correctly after this bug fix.

src/main/scala/com/johnsnowlabs/ml/tensorflow/sentencepiece/SentencePieceWrapper.scala
```diff
@@ -107,6 +107,10 @@ trait ReadSentencePieceModel {
 
     val byteArray = Files.readAllBytes(Paths.get(sppModelFilePath.toString))
     val sppWrapper = new SentencePieceWrapper(byteArray)
+
+    // Delete tmp dir
+    FileUtils.deleteDirectory(new File(tmpFolder))
```
danilojsl: I think we should also cleanup in case of failures. If `copyToLocalFile`, `readAllBytes`, or wrapper creation fails it never reaches this line. We could refactor to something like this:
```scala
try {
  val source = new Path(path, filename)
  val localDir = new Path(tmpFolder)

  fs.copyToLocalFile(source, localDir)

  val localFile = Paths.get(tmpFolder, filename)
  val bytes = Files.readAllBytes(localFile)

  new SentencePieceWrapper(bytes)
} finally {
  FileUtils.deleteDirectory(new File(tmpFolder))
} catch {
 case_: Exception => //ignore cleanup failure
}
```

src/main/scala/com/johnsnowlabs/ml/onnx/OnnxSerializeModel.scala
```diff
@@ -169,6 +179,9 @@ trait ReadOnnxModel {
       (modelName, onnxWrapper)
     }).toMap
 
+    // delete tmp
+    deleteTmpDirectory(tmpFolder.get)
```
danilojsl: Same suggestion than in `SentencePieceWrapper`

danilojsl:
@copilot resolve the merge conflicts in this pull request

PR #14793, opened 2026-07-08
AutoGGUFVisionModel: batch image decoding across parallel slots

C-K-Loan:
## What
`AutoGGUFVisionModel.batchAnnotate` decoded a batch with **one `completeImage()` call per image** (serial, single-stream) — leaving the parallel slots from `setParallel(getBatchSize)` idle. This submits the whole batch in **one** `LlamaExtensions.multiCompleteImage()` call so images decode across those slots. ~16-line change, error handling + `processCompletions` preserved.

## Depends on
⛔ **jsl-llama.cpp#8** (adds `LlamaExtensions.multiCompleteImage`) must be merged + released — this won't compile against the current `jsl-llamacpp` 2.0.3. Draft until then; bump `llamaCppVersion` when the release lands.

## Verified (locally, before opening)
Built spark-nlp under **JDK 8** against a `jsl-llamacpp-gpu` jar patched with `multiCompleteImage`, on a T4:
```
slow:testOnly *AutoGGUFVisionModelTestSpec -- -z "caption the images correctly"
- should caption the images correctly   ✅   (Qwen2.5-VL-3B on CUDA, batchSize=2, 10 images, all captions correct)
Tests: succeeded 5, failed 1
```
(The 1 failure is `should be able to remove thinking tags` — missing `SmolVLM-256M` model file in the local env, unrelated to this change; fails identically on master.)

## Speedup
The parallel-decode win is engine-layer and measured in jsl-llama.cpp#8 (regime-dependent; ~1.8× full-res OCR, more for decode-heavy/small). This PR is the wiring that lets the annotator actually use it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/scala/com/johnsnowlabs/nlp/annotators/seq2seq/AutoGGUFVisionModel.scala
+++ b/src/main/scala/com/johnsnowlabs/nlp/annotators/seq2seq/AutoGGUFVisionModel.scala
@@ -239,23 +239,22 @@ class AutoGGUFVisionModel(override val uid: String)
               .toArray)
       }
 
-      val textAndMeta: Array[(String, Map[String, String])] = prompts
-        .zip(base64EncodedImages)
-        .map { case (prompt, base64Image) =>
-          try {
-            val results = LlamaExtensions.completeImage(
-              model,
-              getInferenceParameters,
-              getSystemPrompt,
-              prompt,
-              base64Image)
-            val resultsCleaned = processCompletions(Array(results)).head
-            (resultsCleaned, Map.empty[String, String])
-          } catch {
-            case e: LlamaException =>
-              logger.error("Error in llama.cpp image batch completion", e)
-              ("", Map("llamacpp_exception" -> e.getMessage))
-          }
+      val textAndMeta: Array[(String, Map[String, String])] =
+        try {
+          // Decode the whole batch in ONE request so images run across the parallel slots configured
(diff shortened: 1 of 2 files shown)


C-K-Loan:
## Added a batch-inference test — verified on GPU

`AutoGGUFVisionModelTestSpec` → **"should batch image inference and be faster than serial on GPU"** (mirrors the perf test in jsl-llama.cpp#8, at the annotator level): builds a serial `batchSize=1` model vs a batched `batchSize=nImages` model, asserts both caption every image correctly **and** — GPU-gated — that the batch is faster (decode overlaps across the parallel slots). `cachePrompt=false` keeps the A/B cache-symmetric; `nCtx` scales with batch size so image tokens don't overflow the per-slot context.

Ran it on a T4 (Qwen2.5-VL-3B, JDK 8, patched `jsl-llamacpp` with `multiCompleteImage`):
```
[VISION BATCH PERF] images=10 | serial(bs=1)=7906ms  batch(bs=10)=6499ms  speedup=1.22x
- should batch image inference and be faster than serial on GPU   ✅ (Tests: succeeded 6, failed 1)
```
Batch beats serial by ~22% with all 10 captions correct in both configs (modest margin because `nPredict=40` is short/encode-dominated — larger for decode-heavy workloads). The 1 unrelated failure is the pre-existing `remove thinking tags` test (needs a `SmolVLM-256M` model file not present locally; fails identically on master).

PR #14830, opened 2026-08-07
[SPARKNLP-1199] implement bge m3 embeddings

ahmedlone127:
BGE-M3 is a multilingual embedding model from BAAI, built on an XLM-RoBERTa-large backbone and released under MIT. The M3 name refers to three properties: multi-functionality (dense, sparse, and multi-vector representations from a single forward pass), multi-linguality (100+ languages, including cross-lingual retrieval), and multi-granularity (inputs up to 8192 tokens).

BGE-M3 lifts all three of those limits at once, and more importantly it adds a sparse output that the library has no encoder-side equivalent for. Dense embeddings capture semantic similarity but can miss exact matches like error codes, SKUs, or proper nouns; sparse lexical weights catch those precisely but are blind to paraphrase. Getting both from one model pass makes hybrid retrieval practical without running two separate models.

This also completes a retrieval path that already exists in the library. `BM25Approach` provides corpus-statistic lexical ranking, `VectorDBConnector` handles dense indexing, and `AutoGGUFReranker` defaults to `bge_reranker_v2_m3`, which is itself trained on `BGE-M3`, so retrieve-then-rerank can now be assembled end to end inside Spark NLP.


src/main/scala/com/johnsnowlabs/ml/ai/BGEM3.scala
```diff
@@ -0,0 +1,360 @@
...
+
+        batch.zipWithIndex.map { case ((_, sentence), idx) =>
+          val metadata = sparseWeightsOpt match {
+            case Some(sparseWeights) =>
+              ListMap(
+                (sentence.metadata.toSeq ++ sparseLexicalWeights(
+                  tokensBatch(idx),
+                  sparseWeights(idx))): _*)
+            case None => sentence.metadata
+          }
```
danilojsl, line 154: **HIGH — MUST FIX**

Sparse lexical weights are merged as top-level metadata keys (`piece → weight.toString`). That can overwrite DocumentAssembler keys such as `sentence` / `id` (the tests already special-case `_STRUCTURAL_KEYS`). `EmbeddingsFinisher` only copies `.embeddings` and then drops the annotation column, so the documented sparse+Finisher path never surfaces this payload.

This is a new public API. Flattening an open SentencePiece vocab is expensive to change after release.

Namespace under one reserved key, e.g. `sparse_weights` → compact JSON, and leave `sentence.metadata` untouched.

**Verification:** a token piece equal to `sentence` must not overwrite `metadata("sentence")`; `select("bge_m3.metadata")` after `setReturnSparseEmbeddings(true)` yields a parseable sparse map.

src/main/scala/com/johnsnowlabs/nlp/embeddings/BGEM3Embeddings.scala
```diff
@@ -0,0 +1,414 @@
...
+  *
+  * val embeddingsFinisher = new EmbeddingsFinisher()
+  *   .setInputCols("bge_m3_embeddings")
+  *   .setOutputCols("finished_embeddings")
+  *   .setOutputAsVector(true)
+  *
+  * val pipeline = new Pipeline().setStages(Array(
+  *   documentAssembler,
+  *   embeddings,
+  *   embeddingsFinisher
```
danilojsl, line 102: **MEDIUM — SHOULD FIX** (same issue in the Python docstring)

This example enables `setReturnSparseEmbeddings(true)` then pipes through `EmbeddingsFinisher`. Finisher reads only `col.embeddings` and, with default `cleanAnnotations=true`, removes the annotation column — so copy-paste users never see sparse weights.

Show sparse via `result.select("bge_m3.metadata")`. Keep Finisher on a dense-only snippet.

src/main/scala/com/johnsnowlabs/ml/ai/BGEM3.scala
```diff
@@ -0,0 +1,360 @@
...
+    val inputs = Map("input_ids" -> tokenTensors, "attention_mask" -> maskTensors).asJava
+
+    try {
+      val results = runner.run(inputs)
+      try {
+        // dense_embedding is already CLS-pooled and L2-normalized in the graph: [batch, dim]
+        val denseTensor = results.get(DenseOutput).get().asInstanceOf[OnnxTensor]
+        val denseDim = denseTensor.getInfo.getShape.last.toInt
+        val dense = denseTensor.getFloatBuffer.array().grouped(denseDim).toArray
+
```
danilojsl, line 206: **MEDIUM — SHOULD FIX**

Sparse output is validated with `expectedSparseWidth` (good). Dense is `grouped(denseDim)` using only `shape.last`. A graph that still emits `[batch, seq, dim]` or a mismatched flat length silently mis-assigns rows.

Assert `flatLength == batch.length * denseDim` and fail with a re-export message, same as sparse.

Same gap in `getEmbeddingsOv` (`denseTensor.data().grouped(denseDim)`).

src/main/scala/com/johnsnowlabs/ml/ai/BGEM3.scala
```diff
@@ -0,0 +1,360 @@
...
+    inferRequest.set_tensor("attention_mask", maskTensors)
+    inferRequest.infer()
+
+    // dense_embedding is already CLS-pooled and L2-normalized in the graph: [batch, dim]
+    val denseTensor = inferRequest.get_tensor(DenseOutput)
+    val denseDim = denseTensor.get_shape().last
+    val dense = denseTensor.data().grouped(denseDim).toArray
+
+    val sparse = if (returnSparse) {
+      val flatSparse = inferRequest.get_tensor(SparseOutput).data()
```
danilojsl, line 249: **MEDIUM — SHOULD FIX**

ONNX checks `results.get(SparseOutput).isPresent` and throws a clear re-export error. OpenVINO calls `get_tensor(SparseOutput)` with no equivalent guard, so a dense-only IR fails opaquely.

Catch a missing output and throw the same `IllegalStateException` as ONNX.

src/test/scala/com/johnsnowlabs/nlp/embeddings/BGEM3EmbeddingsTestSpec.scala
```diff
@@ -0,0 +1,304 @@
...
+        .write
+        .overwrite()
+        .save("./tmp_bge_m3_model")
+    }
+
+    val loadedPipelineModel = PipelineModel.load("./tmp_bge_m3_pipeline")
+    loadedPipelineModel.transform(ddd).select("embeddings.result").show(false)
+
+    val loadedModel = BGEM3Embeddings.load("./tmp_bge_m3_model")
+    assert(loadedModel.getReturnSparseEmbeddings)
```
danilojsl, line 301: **MEDIUM — SHOULD FIX**

Save/load only asserts `getReturnSparseEmbeddings`. It does not compare dense vectors or sparse maps before vs after load, so a broken `readOnnxModel` / spp / external `.onnx_data` path can still pass.

After load, transform the same rows and compare embedding dimension + sparse key set. Even a tiny `loadSavedModel` fixture would catch the data-file suffix.

danilojsl (changes requested):
**Verdict: CHANGES REQUIRED**

Do not merge until finding 1 is addressed. Findings 2–5 should be fixed in the same pass if practical. Finding 6 needs a product answer before `pretrained()` is advertised as the happy path.

### QUESTION — default `pretrained("bge_m3")`

The hub already has `bge_m3` / `xx` as **XlmRoBertaSentenceEmbeddings**. Downloader matches by class so the names can coexist, but `BGEM3Embeddings.pretrained()` 404s until a new `annotator=BGEM3Embeddings` artifact is published, and every SlowTest depends on that. Is that zip shipping with this PR, or should the default name be distinct?

### Looks good

ONNX I/O names match the notebook wrapper; external-data filename matches `OnnxWrapper`; SentencePiece offset / special ids match XlmRoberta; Python Param names and defaults match JVM; `BGEM3Test` covers aggregation and sparse shape (CI-visible).

ahmedlone127:
hey @danilojsl  , I made a commit taking into account all the changes you mentioned as for the pretrained() comments we have a new model for this anntoator already uploaded waiting for you to merge : https://github.com/JohnSnowLabs/spark-nlp/pull/14813/changes/5756b51bff10e108e1bdf7b176a881e96b381de3

ahmedlone127:
hey @danilojsl  , I made a commit taking into account all the changes you mentioned as for the pretrained() comments we have a new model for this anntoator already uploaded waiting for you to merge : https://github.com/JohnSnowLabs/spark-nlp/pull/14813/changes/5756b51bff10e108e1bdf7b176a881e96b381de3

PR #14838, opened 2026-08-15
Sparknlp 1156 fix all compilation warnings

ahmedlone127:
## Description

A clean compile currently emits 68 scalac warnings, and loading the project emits 4 sbt `lintUnused`
warnings before anything is compiled. This clears all 72 without changing runtime behavior.

The scalac warnings break down as:

| Count | Warning | Fix |
|------:|---------|-----|
| 50 | `A try without a catch or finally is equivalent to putting its body in a block` | Unwrapped the redundant *inner* `try` in 19 `ml/ai` classes. Every `catch`/`finally` clause is preserved verbatim. |
| 6 | `non-variable type argument ... is unchecked since it is eliminated by erasure` | `@unchecked` on the type pattern in `HasBatchedAnnotate`, `HasBatchedAnnotateTextImage`, and `Preprocessor` (x4) |
| 4 | `constructor WildcardFileFilter ... is deprecated` | `WildcardFileFilter.builder().setWildcards(...).get()` in `TensorflowWrapper` |
| 2 | `unreachable code` | `HasBatchedAnnotate:54` and `DeBertaForQuestionAnswering:411` |
| 1 | erasure warning on a `val` pattern | `@unchecked` in `AutoGGUFReranker` |
| 1 | `fruitless type test` | `MarianTransformer.getVocabulary` |
| 1 | `comparing values of types BigInt and String using == will always yield false` | `ViTClassifier:179` |
| 1 | `method udf in object functions is deprecated` | `@nowarn("cat=deprecation")` on `PdfToText.transformUDF` |
| 1 | `method getFontIndexAsInt in trait CellStyle is deprecated` | `XlsxParser` -> `getFontIndex` |
| 1 | `inferred existential type ... should be enabled` | `import scala.language.existentials` in `LegacyObjectInputStream` |

Three of these are worth calling out because the obvious fix is wrong:

`PdfToText.transformUDF` keeps the deprecated `udf` overload rather than migrating. The non-deprecated
overload infers its schema by reflection, which cannot produce struct fields named after the
configurable `$(outputCol)` / `$(pageNumCol)`. The deprecated explicit-`DataType` overload is the only
one that works here, so it is suppressed at the definition with a comment rather than replaced.

`MarianTransformer.getVocabulary` has a `fruitless type test` because `$(vocabulary)` is statically
`Array[String]` but is deliberately probed for `java.util.ArrayList`  the existing comment (*"do not
remove or replace with `$(vocabulary)` due to a bug in some models"*) depends on that runtime check
still being emitted. Widening the scrutinee to `Any` silences the fruitless-test warning but trades it
for an *unchecked* warning on the `[String]` type argument, so the method is rewritten as a match on
`java.util.ArrayList[_]`. It also avoids `asScala`, since inferring through the resulting existential
`Buffer[_]` re-triggers the `scala.language.existentials` warning being fixed elsewhere in this PR.

`ViTClassifier` compared a `BigInt` tag index against `index.toString`, which is always false, so the
fallback lookup it guarded could only ever return `"NA"`. The dead chain is removed and `getOrElse("NA")`
is called directly.

No public signature changes, so the Python API is unaffected.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -191,6 +191,14 @@ inConfig(SlowTest)(Defaults.testTasks)
 
 /** Test tagging end */
 
+/** These keys are used via commands (`fast:test`, `slow:test`) or for their side effects rather
+  * than by other settings/tasks, so exclude them from sbt's `lintUnused` check.
+  */
+Global / excludeLintKeys ++= Set(
+  mavenProps,
+  FastTest / configuration,
+  SlowTest / configuration)
+
 /** Enable for debugging */
 (Test / testOptions) += Tests.Argument("-oF")
 
--- a/publish.sbt
+++ b/publish.sbt
@@ -13,7 +13,11 @@ scmInfo := Some(
 licenses += "Apache-2.0" -> url("https://opensource.org/licenses/Apache-2.0")
 
 // Maven Central publishing settings
-ThisBuild / credentials += Credentials(Path.userHome / ".sbt" / "sonatype_central_credentials")
+// Only load the credentials file when it exists to avoid a warning on machines that do not publish.
+ThisBuild / credentials ++= {
+  val credentialsFile = Path.userHome / ".sbt" / "sonatype_central_credentials"
+  if (credentialsFile.exists) Seq(Credentials(credentialsFile)) else Seq.empty
+}
 // Remove all additional repository other than Maven Central from POM
 ThisBuild / pomIncludeRepository := { _ => false }
 ThisBuild / publishMavenStyle := true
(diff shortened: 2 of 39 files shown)


danilojsl (changes requested):
@ahmedlone127 Do we need to apply the same changes for Spark 4.x, or will they require any adjustments to ensure compatibility?

ahmedlone127:
I Haven't tested with spark 4 yet, I will redo these after 7.0.0 are out so we can fix any warnings from thew new features as well

PR #14840, opened 2026-08-19
updating ci/cd pipeline test to support newer python versions

ahmedlone127:
## Description

This PR brings the Python package's declared support in line with what CI actually
verifies, and restructures the build pipeline so that adding a Python version is a
one-line change rather than a copy-pasted job.

**Python 3.12 compatibility**

- Removes `from distutils.version import LooseVersion` in
  `sparknlp/internal/extended_java_wrapper.py`. `distutils` was removed from the
  stdlib in Python 3.12, and this module sits on the normal annotator import path,
  so the package could not be imported on a clean 3.12 environment. Replaced with a
  stdlib tuple comparison, padded to three components so that `"3.3"` still compares
  correctly against `(3, 3, 0)`.
- Converts 8 invalid escape sequences (`"\S+"`, `"\A..."`, `"\z"`) in `Tokenizer`,
  `RegexTokenizer`, `NorvigSweetingApproach` and `SymmetricDeleteApproach` to raw
  strings. These emit `SyntaxWarning` on every import under 3.12 and are scheduled
  to become `SyntaxError`. The resulting string values are byte-identical.
- Removes three dead `sys.version_info[0] == 2` guards.

**Packaging migrated to PEP 621**

- Adds `python/pyproject.toml`; `setup.py` is reduced to a compatibility shim.
- Declares `requires-python = ">=3.8"`. Previously **no `python_requires` existed at
  all**, so pip enforced nothing.
- Classifiers updated to 3.8 through 3.12 (previously 3.6 through 3.9).
- Declares `setuptools; python_version >= "3.12"`. PySpark 3.5.x still imports
  `distutils` in roughly ten of its own modules, so without this `import sparknlp`
  fails inside PySpark on a bare 3.12 virtualenv.
- Removes `setup.cfg`, whose only content was `universal=1`. That produced a
  `py2.py3-none-any` wheel for a package that raises `ImportError` on Python 2; the
  wheel is now correctly tagged `py3-none-any`.
- Sets `namespaces = false` and `include-package-data = false` explicitly. setuptools
  defaults both to `true` under `pyproject.toml`, and the former would have caused
  `python/tensorflow/` to be shipped as a top-level package named `tensorflow`,
  shadowing the real library in every user environment.

**CI pipeline**

- The three hand-copied Spark jobs become a matrix. Each publishes its assembly jar
  as an artifact, and a new 7-cell `python-test` matrix consumes that artifact instead
  of rebuilding the project.
- Adds a `packaging` job that builds the wheel with no PySpark installed, then asserts
  the wheel tag, `Requires-Python`, wheel contents, and that the marker-gated
  `setuptools` dependency resolves on 3.12.
- Removes `**.yml` / `**.yaml` from `paths-ignore`, which previously meant a
  workflow-only change did not trigger the workflow.
- Hardens the JDK install step against the `apt` stalls that repeatedly hung this job
  (`DEBIAN_FRONTEND`, `NEEDRESTART_*`, `DPkg::Lock::Timeout`, per-attempt `timeout`
  and retries), and adds job/step timeouts so a stall fails in minutes instead of
  running to the 6-hour ceiling.
- `python-test` uses `!cancelled()` so that a failure in one Spark build no longer
  skips the entire Python matrix.

**Test harness**

- `python/test/util.py` reads `spark.driver.memory` from
  `SPARKNLP_TEST_DRIVER_MEMORY`, defaulting to the existing `12G`, so the suite can
  run on developer machines with less RAM.

## Motivation and Context

SPARKNLP-1169.

Nothing enforced any claim the package made about Python, and the three sources of
truth disagreed with each other:

- `setup.py` declared no `python_requires` and its classifiers stopped at 3.9, so
  PyPI advertised "up to 3.9" while pip would install on any version, including ones
  where the package could not be imported.
- CI tested Python 3.8 and 3.10 only, and those versions were welded to the Spark
  dimension across three duplicated jobs. There was no floor and no ceiling, just two
  arbitrary points.
- `conda/meta.yaml` pinned `python >=3.7,<3.11`. Unlike the PyPI metadata this was a
  real, enforced cap: it blocked Python 3.11 and above entirely, while
  `docs/en/install.md` simultaneously instructed EMR Serverless users to run 3.11.

The result is one consistent supported range (3.8 through 3.12) across
`requires-python`, trove classifiers, the conda recipe and the documentation, with a
CI matrix that proves it on every push.

## How Has This Been Tested?

**CI** — all 12 jobs green: three Spark builds with the full Scala suite, two
packaging jobs (3.8 and 3.12), and all seven Python matrix cells:

| Spark | PySpark | Python |
|---|---|---|
| 3.3.x | 3.3.1 | 3.8, 3.10 |
| 3.4.x | 3.4.1 | 3.8, 3.11 |
| 3.5.x | 3.5.4 | 3.9, 3.11, 3.12 |

The matrix is deliberately sparse: PySpark 3.3 does not support Python 3.11, and
PySpark 3.4 does not support 3.12, so only combinations PySpark itself supports are
listed.

## Types of changes

## Checklist:

---


ahmedlone127:
@danilojsl 
 **Required status checks.** The nine new job names need adding to branch protection.
   Until then CI proves 3.12 works but nothing prevents it regressing silently.

PR #14848, opened 2026-08-29
feat: add configurable OpenAI API URL to OpenAI annotators

dulcestentaciones2920-debug:
Spark NLP's OpenAICompletion and OpenAIEmbeddings annotators currently hardcode `https://api.openai.com/v1`, so users can only reach the OpenAI API. This PR adds an `apiUrl` parameter to both annotators, backed by a new `spark.jsl.settings.openai.api.url` Spark property that defaults to `https://api.openai.com/v1`. Anyone running Spark NLP against an OpenAI-compatible endpoint — like [OrcaRouter](https://www.orcarouter.ai) — can now point these annotators at it with a single setter or Spark property, without touching application code.

OrcaRouter is an OpenAI-compatible AI gateway built for both models and agents. Like OpenRouter, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding it as a first-class provider means Spark NLP users can use that stack directly, instead of treating OrcaRouter as an anonymous custom base URL. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

The change mirrors how Spark NLP already wires provider configuration through `ConfigHelper`/`ConfigLoader` (e.g. `spark.jsl.settings.openai.api.key`). I verified the new code path against OrcaRouter's live `/v1/embeddings` endpoint (HTTP 200, 3072-dim embedding returned) and kept the default behavior identical to today for anyone not overriding the URL.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.

### DIFF
--- a/python/sparknlp/annotator/openai/openai_completion.py
+++ b/python/sparknlp/annotator/openai/openai_completion.py
@@ -55,6 +55,9 @@ class OpenAICompletion(AnnotatorModel):
       Modify the likelihood of specified tokens appearing in the completion.
    user
       A unique identifier representing your end-user, which can help OpenAI to monitor and detect abuse.
+   apiUrl
+      Base URL of the OpenAI-compatible API. Defaults to the value of spark.jsl.settings.openai.api.url
+      or https://api.openai.com/v1.
 
    Examples
    --------
@@ -163,6 +166,26 @@ class OpenAICompletion(AnnotatorModel):
                  "user",
                  "A unique identifier representing your end-user, which can help OpenAI to monitor and detect abuse.",
                  typeConverter=TypeConverters.toString)
+
+    apiUrl = Param(Params._dummy(),
+                   "apiUrl",
+                   "Base URL of the OpenAI-compatible API. Defaults to the value of spark.jsl.settings.openai.api.url or https://api.openai.com/v1.",
+                   typeConverter=TypeConverters.toString)
+
+    def setApiUrl(self, value):
+        """Sets the base URL of the OpenAI-compatible API.
+
+        By default this is the value of the ``spark.jsl.settings.openai.api.url`` Spark property,
+        or ``https://api.openai.com/v1`` when the property is not set. Pointing it at an
+        OpenAI-compatible gateway (for example OrcaRouter) lets this annotator call any provider
+        that exposes the same API without changing application code.
+
+        Parameters
+        ----------
+        value : str
+           Base URL of the OpenAI-compatible API.
+        """
+        return self._set(apiUrl=value)
 
     def setModel(self, value):
         """Sets model ID of the OpenAI model to use
--- a/python/sparknlp/annotator/openai/openai_embeddings.py
+++ b/python/sparknlp/annotator/openai/openai_embeddings.py
@@ -31,6 +31,9 @@ class OpenAIEmbeddings(AnnotatorModel):
         ID of the OpenAI model to use.
     user
         A unique identifier representing your end-user, which can help OpenAI to monitor and detect abuse.
+    apiUrl
+        Base URL of the OpenAI-compatible API. Defaults to the value of spark.jsl.settings.openai.api.url
+        or https://api.openai.com/v1.
 
     Examples
     --------
@@ -77,6 +80,26 @@ class OpenAIEmbeddings(AnnotatorModel):
                  "user",
                  "A unique identifier representing your end-user, which can help OpenAI to monitor and detect abuse.",
                  typeConverter=TypeConverters.toString)
+
+    apiUrl = Param(Params._dummy(),
+                   "apiUrl",
+                   "Base URL of the OpenAI-compatible API. Defaults to the value of spark.jsl.settings.openai.api.url or https://api.openai.com/v1.",
+                   typeConverter=TypeConverters.toString)
+
+    def setApiUrl(self, value):
+        """Sets the base URL of the OpenAI-compatible API.
+
+        By default this is the value of the ``spark.jsl.settings.openai.api.url`` Spark property,
+        or ``https://api.openai.com/v1`` when the property is not set. Pointing it at an
+        OpenAI-compatible gateway (for example OrcaRouter) lets this annotator call any provider
+        that exposes the same API without changing application code.
+
+        Parameters
+        ----------
+        value : str
+           Base URL of the OpenAI-compatible API.
+        """
+        return self._set(apiUrl=value)
 
     def setModel(self, value):
         """Sets model ID of the OpenAI model to use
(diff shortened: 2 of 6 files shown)


danilojsl (commented):
Please review the comments below

src/main/scala/com/johnsnowlabs/ml/ai/OpenAICompletion.scala
```diff
@@ -110,6 +110,18 @@ class OpenAICompletion(override val uid: String)
...
   def setModel(value: String): this.type = set(model, value)
 
+  val apiUrl = new Param[String](
+    this,
+    "apiUrl",
+    "Base URL of the OpenAI-compatible API. Defaults to the value of spark.jsl.settings.openai.api.url or https://api.openai.com/v1.")
+
+  def setApiUrl(value: String): this.type = set(apiUrl, value)
+
+  def getApiUrlOrDefault: String = {
```
danilojsl, line 120: HIGH — this runs on executors, not the driver.

`annotate` / `post` call `getApiUrlOrDefault`. When `apiUrl` is unset it reads `ConfigLoader.getConfigStringValue`, which snapshots `SparkSession.conf` via `ResourceHelper.spark`. That is a driver-side lazy map; it is not shipped with the stage.

`setApiUrl` works because the Param serializes. `spark.jsl.settings.openai.api.url` alone will still hit `api.openai.com` on a cluster. `local[*]` hides this.

In `beforeAnnotate` (already used to copy the API key on the driver), if `!isSet(apiUrl)` then `set(apiUrl, ConfigLoader.getConfigStringValue(ConfigHelper.openAIApiUrl))` and use `$(apiUrl)` in `post`.

Add a FastTest (no network) for default vs conf vs setter vs trailing slash.

src/main/scala/com/johnsnowlabs/ml/ai/OpenAICompletion.scala
```diff
@@ -286,7 +298,7 @@ class OpenAICompletion(override val uid: String)
         bestOfJson,
         logitBiasJson,
         userJson))
-    val openAIUrlCompletion = "https://api.openai.com/v1/completions"
+    val openAIUrlCompletion = getApiUrlOrDefault + "/completions"
```
danilojsl, line 301: HIGH — join the path only after `apiUrl` is a Param on the stage, and strip a trailing `/`.

`https://api.openai.com/v1/` + `"/completions"` becomes `…/v1//completions`. After resolving the URL in `beforeAnnotate`, use `$(apiUrl).stripSuffix("/") + "/completions"`.

src/main/scala/com/johnsnowlabs/ml/ai/OpenAIEmbeddings.scala
```diff
@@ -108,7 +108,17 @@ class OpenAIEmbeddings(override val uid: String)
...
 
-  private val openAIUrlEmbeddings = "https://api.openai.com/v1/embeddings"
+  val apiUrl = new Param[String](
+    this,
+    "apiUrl",
+    "Base URL of the OpenAI-compatible API. Defaults to the value of spark.jsl.settings.openai.api.url or https://api.openai.com/v1.")
+
+  def setApiUrl(value: String): this.type = set(apiUrl, value)
+
+  def getApiUrlOrDefault: String = {
```
danilojsl, line 118: HIGH — same executor/ConfigLoader problem as `OpenAICompletion.getApiUrlOrDefault`.

`post()` runs in `annotate` on executors. Unset `apiUrl` will not see `spark.jsl.settings.openai.api.url` on a cluster.

In this class’s `beforeAnnotate` (already loads the API key), if `!isSet(apiUrl)` set it from ConfigLoader so the Param serializes. Then `post` can use `$(apiUrl)`.

Add the same FastTest cases (no network).

src/main/scala/com/johnsnowlabs/ml/ai/OpenAIEmbeddings.scala
```diff
@@ -138,7 +148,7 @@ class OpenAIEmbeddings(override val uid: String)
   }
 
   private def post(jsonBody: String): Array[Float] = {
-    val httpPost = new HttpPost(openAIUrlEmbeddings)
+    val httpPost = new HttpPost(getApiUrlOrDefault + "/embeddings")
```
danilojsl, line 151: HIGH — same join as completions: a trailing slash doubles the path.

Use the Param value after `beforeAnnotate`, `stripSuffix("/") + "/embeddings"`.

python/sparknlp/annotator/openai/openai_completion.py
```diff
@@ -164,6 +167,26 @@ class OpenAICompletion(AnnotatorModel):
...
+                   "apiUrl",
+                   "Base URL of the OpenAI-compatible API. Defaults to the value of spark.jsl.settings.openai.api.url or https://api.openai.com/v1.",
+                   typeConverter=TypeConverters.toString)
+
+    def setApiUrl(self, value):
+        """Sets the base URL of the OpenAI-compatible API.
+
+        By default this is the value of the ``spark.jsl.settings.openai.api.url`` Spark property,
+        or ``https://api.openai.com/v1`` when the property is not set. Pointing it at an
+        OpenAI-compatible gateway (for example OrcaRouter) lets this annotator call any provider
```
danilojsl, line 180: SHOULD FIX — keep this Param generic.

`apiUrl` is any OpenAI-compatible base URL (`scheme + host + /v1`). Naming OrcaRouter in the library docstring is vendor advertising, not API contract. Drop the product name; a generic gateway example is enough.

python/sparknlp/annotator/openai/openai_embeddings.py
```diff
@@ -78,6 +81,26 @@ class OpenAIEmbeddings(AnnotatorModel):
...
+                   "apiUrl",
+                   "Base URL of the OpenAI-compatible API. Defaults to the value of spark.jsl.settings.openai.api.url or https://api.openai.com/v1.",
+                   typeConverter=TypeConverters.toString)
+
+    def setApiUrl(self, value):
+        """Sets the base URL of the OpenAI-compatible API.
+
+        By default this is the value of the ``spark.jsl.settings.openai.api.url`` Spark property,
+        or ``https://api.openai.com/v1`` when the property is not set. Pointing it at an
+        OpenAI-compatible gateway (for example OrcaRouter) lets this annotator call any provider
```
danilojsl, line 94: SHOULD FIX — same as `OpenAICompletion.setApiUrl`: remove the OrcaRouter name from the public Param docs. Describe an OpenAI-compatible base URL only.

PR #14849, opened 2026-08-29
[SPARKNLP-936] Add built-in benchmarking API (throughput + accuracy)

AbdullahMubeenAnwar:
This PR adds a public `Benchmark` API, in Scala as `com.johnsnowlabs.nlp.benchmark.Benchmark` and in Python as `sparknlp.benchmark.Benchmark`, that measures throughput and accuracy for any already fitted `PipelineModel` against data the caller supplies.

## `throughput`

Runs warmup passes followed by several timed trials over the given data and reports a rate (items per second) with a 95% confidence interval, for every annotation type the pipeline actually produces. It reads output types straight from each column's existing `annotatorType` schema metadata, so it works unmodified on any pipeline.

```scala
val report = Benchmark.throughput(pipelineModel, data)
println(report)
```
```python
report = Benchmark.throughput(pipeline_model, data)
print(report)
```

For example:

```text
Throughput over 5 trial(s), mean elapsed 1.842 sec/trial
  token                1,204.3 ± 22.1 items/sec (type: token, n=6021)
  named_entity         1,198.7 ± 19.4 items/sec (type: named_entity, n=6021)
```

## `evaluate`

Scores accuracy for twelve task types (NER, POS, dependency parsing, classification, spell checking, language detection, word segmentation, image classification, question answering, speech recognition, translation, and summarization) against gold data the caller supplies.

```scala
val report = Benchmark.evaluate(pipelineModel, goldData, BenchmarkTask.NER)
println(report)
```
```python
report = Benchmark.evaluate(pipeline_model, gold_data, task="ner")
print(report)
```

The output column to score comes from the pipeline's own annotator types. Several stages often emit the same type, for example a `DocumentAssembler`'s output shares `document` with a translation or summary stage, so `evaluate` scores the last one (the actual prediction) and names it in the report:

```text
Translation accuracy (n=200, scored: translation): bleu=0.3421, ...
```

Pass `predictedCol`/`predicted_col` to override that when a pipeline's stage order puts the column you want somewhere other than last.

These route through five comparison engines: per item label accuracy and top k accuracy built on Spark's own `MulticlassMetrics` so nothing needs to be collected to the driver, span level F1 for NER and word segmentation, head and label accuracy for dependency parsing, and text similarity metrics (BLEU, ROUGE, WER, SQuAD style exact match/F1) that reimplement the same defaults as `sacrebleu`, `rouge_score`, and `jiwer` so Scala and Python report matching numbers on the same input.

## Datasets

The library doesn't auto download any dataset. Both `data` and `goldData`/`gold_data` are required, since scoring against a canned public dataset answers a different question than scoring against the caller's actual data. `examples/python/benchmarks/` has thirteen Colab ready notebooks: one for `throughput`, and one for each of the twelve `evaluate` task types, each downloading a specific recommended public dataset and running it end to end through the new API.

## Dependencies

`sacrebleu`, `rouge-score`, and `jiwer` are new optional Python dependencies, installed via `pip install spark-nlp[benchmark]`.


AbdullahMubeenAnwar:
TODO: add docs

python/sparknlp/benchmark.py
```diff
@@ -0,0 +1,840 @@
...
+        return (best_em, best_f1, 1)
+
+    em_sum, f1_sum, n = pairs.map(row_scores).fold(
+        (0.0, 0.0, 0), lambda a, b: (a[0] + b[0], a[1] + b[1], a[2] + b[2]))
+    if n == 0:
+        return {}, 0
+    return {"exactMatch": em_sum / n, "f1": f1_sum / n}, n
+
+
+def _wer(pairs):
```
danilojsl, line 748: HIGH — this Python WER path is not the JVM engine.

`jiwer.process_words` runs inside the Spark map, so tokenization follows jiwer (tabs are not word breaks). Scala `werTokenize` uses `split("\\s+")`, so the same gold string `"the\\tcat sat  on the mat"` is WER 0.4 here and WER 0 in Scala.

`jiwer` must also be importable on every executor, not only the driver.

Please score ASR through `com.johnsnowlabs.nlp.benchmark.Benchmark` (Py4J), or tokenize identically to Scala and keep the metric fully distributed without a worker-side extra.

src/main/scala/com/johnsnowlabs/nlp/benchmark/engines/TextSimilarityEngine.scala
```diff
@@ -0,0 +1,329 @@
...
+    // period/comma split only when not flanked by a digit on the relevant side
+    line = line.replaceAll("([^0-9])([\\.,])", "$1 $2 ")
+    line = line.replaceAll("([\\.,])([^0-9])", " $1 $2")
+    // hyphen split only when preceded by a digit
+    line = line.replaceAll("([0-9])(-)", "$1 $2 ")
+    line.split("\\s+").filter(_.nonEmpty)
+  }
+
+  // jiwer's default WER tokenization is plain whitespace splitting.
+  private def werTokenize(text: String): Array[String] =
```
danilojsl, line 64: HIGH — this is the Scala half of the WER mismatch.

`trim.split("\\s+")` treats tab/newline as separators. Python `_wer` uses jiwer’s default pipeline, which does not. The public API claims matching numbers on the same input.

Keep one implementation (Python should call this JVM engine), or change this tokenizer only if both languages and the tests agree on the same rule.

python/test/benchmark_test.py
```diff
@@ -0,0 +1,341 @@
...
+
+    def test_wer_matches_scala_side_computation(self):
+        pytest.importorskip("jiwer")
+        pairs = self.sc.parallelize([("the cat sat on mat", "the cat sat on the mat")])
+        overall, support = _wer(pairs)
+
+        self.assertEqual(support, 1)
+        self.assertAlmostEqual(overall["wer"], 1.0 / 6)
+
+    def test_wer_matches_jiwer_itself_on_a_whitespace_dirty_reference(self):
```
danilojsl: HIGH — this fixture locks Python to jiwer, not to Scala.

It asserts `wer == 0.4` because jiwer keeps `"the\\tcat"` as one token. The same pair on `TextSimilarityEngine` with `\\s+` splitting is a perfect match (WER 0).

CI also `importorskip("jiwer")`, so this never runs on the Spark 3.x pytest jobs (they install only pyspark/numpy/pytest).

After unifying the engines, this case should assert the shared contract and run without an optional extra.

python/sparknlp/benchmark.py
```diff
@@ -0,0 +1,840 @@
...
+        # this call passes exactly one row.
+        return (edits, len(result.references[0]), 1)
+
+    total_edits, total_words, n = pairs.map(row_stats).fold(
+        (0, 0, 0), lambda a, b: (a[0] + b[0], a[1] + b[1], a[2] + b[2]))
+    score = total_edits / total_words if total_words else 0.0
+    return {"wer": score}, n
+
+
+def _bleu(pairs):
```
danilojsl, line 790: HIGH — Python BLEU is a different API than Scala BLEU.

`list(pairs.toLocalIterator())` materializes the whole hypothesis/reference corpus on the driver, then `sacrebleu.corpus_bleu` returns only `bleu`. Scala `TextSimilarityEngine.bleu` stays distributed and also reports `precision1..4` and `brevityPenalty`.

Please call the JVM `Benchmark.evaluate` from Python (no driver collect, same `overall` keys), or match that contract without collecting.

python/sparknlp/benchmark.py
```diff
@@ -0,0 +1,840 @@
...
+    rows = list(pairs.toLocalIterator())
+    if not rows:
+        return {}, 0
+    hypotheses = [p for p, _ in rows]
+    references = [[g for _, g in rows]]
+    result = sacrebleu.corpus_bleu(hypotheses, references)
+    return {"bleu": result.score / 100.0}, len(rows)
+
+
+def _rouge(pairs):
```
danilojsl: HIGH — ROUGE runs `rouge_score` inside `mapPartitions`.

That package has to exist on every executor. `pip install spark-nlp[benchmark]` only installs it on the client. Local-mode tests hide this; a YARN/K8s worker will `ImportError` on the first partition.

Scala already implements ROUGE in the JVM. Route Python `task="summarization"` through that instead of a worker-side extra.

python/setup.py
```diff
@@ -141,6 +141,13 @@
 
     packages=find_packages(exclude=['test*', 'tmp*']),
 
-    include_package_data=False  # Needed to install jar file
+    include_package_data=False,  # Needed to install jar file
+
+    # Optional dependencies for sparknlp.benchmark's text-similarity metrics (BLEU, ROUGE,
+    # WER) -- install with `pip install spark-nlp[benchmark]`. Kept out of the base install
+    # since most users never call these.
+    extras_require={
```
danilojsl: HIGH — `extras_require['benchmark']` does not match how `_wer` / `_rouge` run.

Those imports happen in executor closures. Installing the extra on the driver is not sufficient on a cluster. Translation/summarization/ASR notebooks never install this extra at all, and README does not mention it.

If Python calls the JVM Benchmark API, this extra can go away. If it stays, document that the packages must be on every executor, install them in the three notebooks, and add them to the CI pytest jobs so `importorskip` stops hiding the path.

examples/python/benchmarks/Benchmark_Translation.ipynb
```diff
@@ -0,0 +1,330 @@
...
+     "text": [
+      "translation accuracy (n=200): bleu=0.5460"
+     ]
+    }
+   ],
+   "source": [
+    "# Both DocumentAssembler and MarianTransformer emit annotatorType \"document\". Benchmark scores\n",
+    "# the last one -- the translation -- and names it in the report, so you can confirm at a glance\n",
+    "# that the score isn't accidentally measuring the untranslated source text.\n",
+    "report = Benchmark.evaluate(pipeline_model, gold_data, task=\"translation\", label_col=\"label\")\n",
```
danilojsl, line 257: HIGH — this cell will `ImportError` on a fresh Colab.

`Benchmark.evaluate(..., task="translation")` imports `sacrebleu` in `_bleu`. This notebook never `pip install`s `spark-nlp[benchmark]` or `sacrebleu` (only the colab.sh Spark NLP install).

Either drop the Python extra by calling the JVM engine, or install the extra in Setup before this cell.

danilojsl (commented):
Check the review comments below

examples/python/benchmarks/Benchmark_Summarization.ipynb
```diff
@@ -0,0 +1,361 @@
...
+     "name": "stdout",
+     "text": [
+      "summarization accuracy (n=8): rouge1_f1=0.2825, rouge1_precision=0.2214, rouge1_recall=0.4045, rouge2_f1=0.1172, rouge2_precision=0.0921, rouge2_recall=0.1646, rougeL_f1=0.2358, rougeL_precision=0.1861, rougeL_recall=0.3315"
+     ]
+    }
+   ],
+   "source": [
+    "# DocumentAssembler and BartTransformer both emit annotatorType \"document\"; Benchmark scores the\n",
+    "# last one (the summary) and names it in the report.\n",
+    "report = Benchmark.evaluate(pipeline_model, gold_data, task=\"summarization\", label_col=\"label\")\n",
```
danilojsl, line 302: HIGH — same gap as Translation, for ROUGE.

`task="summarization"` needs `rouge-score` on the Python workers. This notebook does not install it. `evaluate` will fail on Colab, and on a cluster even after a driver-only `pip install`.

Call the JVM engine, or install the extra in Setup and state that workers need it too.

examples/python/benchmarks/Benchmark_SpeechRecognition.ipynb
```diff
@@ -0,0 +1,360 @@
...
+      "Apache Spark version: 3.4.4"
+     ]
+    }
+   ],
+   "source": [
+    "import sparknlp\n",
+    "spark = sparknlp.start()\n",
+    "print(\"Spark NLP version:\", sparknlp.version())\n",
+    "print(\"Apache Spark version:\", spark.version)\n",
+    "!pip install -q soundfile\n",
```
danilojsl, line 142: HIGH — `soundfile` is installed here, but ASR scoring needs `jiwer` on executors (`_wer`).

This notebook never installs `jiwer` / `spark-nlp[benchmark]`. `evaluate(task="speechrecognition")` will fail unless that extra is already on every worker.

Prefer JVM WER (no extra). If jiwer stays, install it in this cell and keep tokenization in lockstep with Scala.

src/main/scala/com/johnsnowlabs/nlp/benchmark/Benchmark.scala
```diff
@@ -0,0 +1,483 @@
...
+    */
+  def throughput(
+      pipelineModel: PipelineModel,
+      data: DataFrame,
+      textCol: String = "text",
+      warmupRuns: Int = 1,
+      trials: Int = 5): ThroughputReport = {
+    require(warmupRuns >= 0, "warmupRuns must be >= 0")
+    require(trials >= 1, "trials must be >= 1")
+    require(data.schema.fieldNames.contains(textCol), s"data must contain a '$textCol' column")
```
danilojsl: SHOULD FIX — `textCol` is required but never read.

This is only an existence check. `transform(data)` uses whatever columns the pipeline stages declare. Image/audio frames have no `text`, so `throughput(vitModel, imageDf)` throws unless the caller passes a dummy name.

Drop the check, or make `textCol` actually select the input column.

src/main/scala/com/johnsnowlabs/nlp/benchmark/Benchmark.scala
```diff
@@ -0,0 +1,483 @@
...
+      textCol: String = "text",
+      labelCol: String = "label",
+      predictedCol: Option[String] = None,
+      predictedLabeledDependencyCol: Option[String] = None,
+      topK: Int = 5): AccuracyReport = {
+
+    require(topK >= 1, "topK must be >= 1")
+    require(
+      goldData.schema.fieldNames.contains(textCol),
+      s"goldData must contain a '$textCol' column")
```
danilojsl: SHOULD FIX — same unused `textCol` gate on `evaluate`.

ImageClassification documents `textCol` as the image column; SpeechRecognition gold typically has `audio` + `label` and no `text`. The column is still not passed into `transform`.

Do not require a column the scoring path does not use.

python/sparknlp/benchmark.py
```diff
@@ -0,0 +1,840 @@
...
+        again afterwards) -- otherwise a ``data`` sourced from an expensive or non-deterministic
+        upstream read (a file scan, a ``.sample()``) would have that cost repeated on every pass,
+        contaminating the measured rate and, for a non-deterministic source, varying the row
+        count between trials.
+        """
+        if warmup_runs < 0:
+            raise ValueError("warmup_runs must be >= 0")
+        if trials < 1:
+            raise ValueError("trials must be >= 1")
+        if text_col not in data.columns:
```
danilojsl: SHOULD FIX — Python `throughput` copies the unused `text_col` requirement.

If the DataFrame only has `image` / `audio_content`, this raises even though the pipeline never reads `text`. Mirror whatever Scala does after that check is removed (or actually use the column).

python/sparknlp/benchmark.py
```diff
@@ -0,0 +1,840 @@
...
+
+        ``top_k`` sizes the top-k window for ``imageclassification``'s extra
+        ``top{top_k}Accuracy`` metric; every other task ignores it.
+        """
+        task = task.lower()
+        if task not in SUPPORTED_TASKS:
+            raise ValueError(f"Unknown benchmark task '{task}'. Supported: {sorted(SUPPORTED_TASKS)}")
+        if top_k < 1:
+            raise ValueError("top_k must be >= 1")
+        if text_col not in gold_data.columns:
```
danilojsl: SHOULD FIX — Python `evaluate` also requires `text_col` to exist and then ignores it.

The notebooks work around this (`text_col="image"`, `text_col="audio_content"`, `text_col="context"`). That is a dummy override for a column scoring does not read.

Drop the check unless `text_col` becomes a real input selector.

AbdullahMubeenAnwar:
wer/bleu/rouge in python are now pure python ports of the same algorithms already in TextSimilarityEngine.scala instead of calling jiwer/sacrebleu/rouge_score, so both languages tokenize identically and nothing needs installing on the executors anymore. dropped the extras_require block from setup.py since there's nothing left to install. also dropped the textCol/text_col existence checks in throughput/evaluate on both languages since the column was validated but never actually read afterward. updated the wer test that was asserting jiwer's tab-handling quirk, and added real partial-overlap/punctuation fixtures for rouge that were missing before (it only had an identical-strings test). ran all three notebooks that were flagged end to end on a fresh colab VM with none of those packages installed to confirm.

PR #14850, opened 2026-08-30
Fix stray spacing in QA span detokenization

AbdullahMubeenAnwar:
predictSpan in XXXForClassification.scala, RoBertaClassification.scala, MPNetClassification.scala, and ZeroShotNerClassification.scala all used the same inline logic to join WordPieces back into text, and none of them undid HF's clean_up_tokenization_spaces step. So QA answers came out like "Levi ' s Stadium" instead of "Levi's Stadium".

Moved that logic into one shared function on XXXForClassification (joinWordPieces + cleanUpTokenizationSpaces, based on HF's actual rules) and pointed all four call sites at it, so the same code isn't copied four times anymore.

### DIFF
--- a/src/main/scala/com/johnsnowlabs/nlp/annotators/coref/SpanBertCorefModel.scala
+++ b/src/main/scala/com/johnsnowlabs/nlp/annotators/coref/SpanBertCorefModel.scala
@@ -15,7 +15,7 @@
  */
 package com.johnsnowlabs.nlp.annotators.coref
 
-import com.johnsnowlabs.ml.ai.SpanBertCoref
+import com.johnsnowlabs.ml.ai.{MergeTokenStrategy, SpanBertCoref, XXXForClassification}
 import com.johnsnowlabs.ml.tensorflow.{
   ReadTensorflowModel,
   TensorflowWrapper,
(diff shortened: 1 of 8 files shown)


src/main/scala/com/johnsnowlabs/nlp/annotators/coref/SpanBertCorefModel.scala
```diff
@@ -347,23 +347,17 @@ class SpanBertCorefModel(override val uid: String)
...
-        .map(x => (if (x._1.isWordStart) " " else "") + x._1.wordpiece.replaceFirst("##", ""))
-        .mkString("")
-        .trim
+      // WordpieceEncoder-tokenized (see `tokenizedSentences`, built with `WordpieceEncoder.encode`
+      // above), so the vocab merge strategy applies -- same reconstruction RoBertaClassification/
+      // MPNetClassification/XXXForClassification use, undoing the extra spaces a naive
+      // word-start-token join introduces around punctuation and contractions (e.g.
+      // "Levi ' s Stadium" -> "Levi's Stadium").
+      val clusterHeadSpanText =
+        XXXForClassification.joinWordPieces(clusterHeadSpan.map(_._1), MergeTokenStrategy.vocab)
```
danilojsl, line 360: SHOULD FIX — this is not the same reconstruction coref used before.

The old join was every piece in the span:
`(if (isWordStart) " " else "") + wordpiece.replaceFirst("##", "")`

`joinWordPieces(..., vocab)` keeps only `isWordStart` pieces and reads `.token` (the whole original word). Word-aligned mentions still work (`Den`+`##ver` → `"Denver"`) and contractions get cleanup. A span that starts on a continuation piece does not: `[##ver]` → `""` instead of `"ver"`.

Apply `cleanUpTokenizationSpaces` to the previous `##`-strip join (or a join that still emits continuation `wordpiece`s). Do not switch coref onto the QA vocab filter.

Verify with a span whose first `TokenPiece` has `isWordStart = false` and assert the mention is not empty.

danilojsl (commented):
Please review the comments below

src/main/scala/com/johnsnowlabs/nlp/annotators/coref/SpanBertCorefModel.scala
```diff
@@ -380,10 +374,7 @@ class SpanBertCorefModel(override val uid: String)
           annotatorType = AnnotatorType.DEPENDENCY,
           begin = span.head._1.begin,
           end = span.last._1.end,
-          result = span
-            .map(x => (if (x._1.isWordStart) " " else "") + x._1.wordpiece.replaceFirst("##", ""))
-            .mkString("")
-            .trim,
+          result = XXXForClassification.joinWordPieces(span.map(_._1), MergeTokenStrategy.vocab),
```
danilojsl, line 377: SHOULD FIX — same vocab-filter change as the cluster-head join above.

Non-head mentions also used to include continuation `wordpiece`s (`##` stripped). This drops them. Keep the old piece join here too, then run `cleanUpTokenizationSpaces`.

src/test/scala/com/johnsnowlabs/nlp/e2e/QaSpacingE2ESpec.scala
```diff
@@ -0,0 +1,49 @@
...
+      .pretrained("roberta_base_qa_squad2", "en")
+      .setInputCols(Array("document_question", "document_context"))
+      .setOutputCol("answer")
+      .setCaseSensitive(true)
+
+    val pipeline = new Pipeline().setStages(Array(documentAssembler, questionAnswering))
+    val result = pipeline.fit(data).transform(data)
+
+    val answers = result.selectExpr("explode(answer.result) as r").as[String].collect()
+    answers.foreach(a => println(s"ANSWER=[$a]"))
```
danilojsl, line 41: SUGGESTION — this SlowTest does not actually catch the second-commit `'t` case, and it prints.

`"Don 't be evil"` contains neither `" ' "`, `" 's"`, nor `" n't"`, so the asserts below would pass. Drop `println`, forbid `" 't"` as well (or assert the expected answer strings).

PR #14860, opened 2026-09-04
Update documentation for Spark 4

AbdullahMubeenAnwar:


danilojsl:
Thanks @AbdullahMubeenAnwar 
@ahmedlone127 There are some conflicts that need to be resolved. Please keep in mind that we will continue supporting both Spark 3.x / Scala 2.12 and Spark 4.x / Scala 2.13.

Because of that, we shouldn’t remove the existing references or setup for Spark 3.x / Scala 2.12. Instead, we should complement them by adding the corresponding Spark 4.x / Scala 2.13 references, configuration, or setup wherever needed. The goal is to preserve compatibility with both versions rather than replacing one with the other.

PR #14864, opened 2026-09-04
Sparknlp 1430 create script for colab saprk 3.x

ahmedlone127:
this pr introduces a colab installation script to downgrade python versions on colab to 3.11 to support backwards compatibility with spark 3.x

### DIFF
--- a/scripts/colab_setup_python.sh
+++ b/scripts/colab_setup_python.sh
@@ -1,0 +1,294 @@
+#!/bin/bash
+#
+# Google Colab setup for Spark NLP + PySpark, including a downgrade of the
+# notebook kernel's Python version.
+#
+# WHY THIS EXISTS
+#   Colab's default runtime now ships Python 3.13. No PySpark 3.x release
+#   supports it - pyspark 3.4.x and 3.5.x both declare support only through
+#   Python 3.11 - and Spark NLP 6.x runs on Apache Spark 3.0.x-3.5.x only, so
+#   "just use Spark 4" is not an option either. scripts/colab_setup.sh installs
+#   PySpark against whatever Python Colab gives it, which no longer works.
+#
+#   This script does everything colab_setup.sh does (Java, optional GPU cuDNN,
+#   pyspark + spark-nlp + findspark) and additionally replaces the interpreter
+#   behind Colab's default "Python 3" kernel with an older one, so the cells
+#   you run after a restart execute on a Python that PySpark supports.
+#
+# HOW IT WORKS
+#   Miniconda for the target Python is installed over /usr/local, the packages
+#   Colab's frontend needs to talk to a kernel (jupyter, google-colab,
+#   traitlets) are reinstalled there, and a Jupyter kernelspec named "python3"
+#   is written to ~/.local/share/jupyter/kernels. That user-level kernelspec
+#   shadows Colab's system one, so restarting the runtime relaunches the
+#   default kernel on the new interpreter - no notebook metadata edits needed.
+#   Technique adapted from https://github.com/j3soon/colab-python-version
+#
+# USAGE (in a Colab cell)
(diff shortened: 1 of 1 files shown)


scripts/colab_setup_python.sh
```diff
@@ -0,0 +1,282 @@
...
+PYSPARK="3.4.4"
+PYTHON_VERSION="3.11"
+MINICONDA_BUILD="26.7.1-1"
+TRAITLETS="5.5.0"
+GPU="false"
+FORCE="false"
+
+usage() {
+  awk '/^# USAGE/,/^#   -h  help/' "$0" | cut -c3-
+}
```
danilojsl, line 40: **`usage()` no longer prints anything.**

This function extracts comment lines between `# USAGE` and `#   -h  help`. The trim commit removed that header, so `bash scripts/colab_setup_python.sh -h` exits 0 with empty stdout. Invalid flags hit the same empty `usage`.

Restore a short usage string in the script body (not only comments), including the Colab one-liner and the real `-v/-s/-p/-m/-g/-f` flags.

Verified locally: `bash -n` is clean; `-h` prints nothing.

scripts/colab_setup_python.sh
```diff
@@ -0,0 +1,282 @@
...
+  echo "       Pick an existing build with -m (see https://repo.anaconda.com/miniconda/)." >&2
+  exit 1
+fi
+
+echo "[2/7] Java (Spark 3.x requires Java 8, 11 or 17)..."
+JAVA_HOME="/usr/lib/jvm/java-11-openjdk-amd64"
+if [ ! -d "$JAVA_HOME" ]; then
+  echo "       Installing OpenJDK 11..."
+  apt-get update -qq >/dev/null 2>&1 || true
+  apt-get install -y -qq openjdk-11-jdk >/dev/null 2>&1 || true
```
danilojsl: **Do not swallow the OpenJDK 11 install.**

`apt-get install … || true` plus `>/dev/null` hides the failure. Colab already has Java 21, which Spark 3.x rejects, so this install is the actual fix. If apt fails, the script later only says no Spark-compatible JDK was found, with no apt log.

Drop `|| true` on the install, keep apt output on failure, and use the 17/8 fallback only when those trees already exist on disk.

danilojsl (commented):
Two review comments on `scripts/colab_setup_python.sh` (help text and JDK install error handling). Not requesting changes on the rest of the script from this review.

scripts/colab_setup_python.sh
```diff
@@ -0,0 +1,282 @@
...
+fi
+
+if [[ "$PREFLIGHT_OK" != "true" ]]; then
+  echo
+  echo "------------------------------------------------------------------"
+  echo "PRE-FLIGHT CHECK FAILED - the default kernel was NOT switched."
+  echo
+  echo "Python ${PYTHON_VERSION} cannot start a Colab kernel in this runtime."
+  echo "Nothing that would stop the runtime from reconnecting has been written,"
+  echo "so this session is still usable."
```
danilojsl: **This banner is not about the Spark NLP wheel version (6.x vs 7.0.0).**

It only runs after pip has already succeeded. An unpublished `spark-nlp==7.0.0` fails earlier at step 5 and hits the ERR trap, which correctly says `/usr/local` may be partly rewritten.

By the time we get here, Miniconda has already run `bash … -p /usr/local`. Pre-flight can still fail for Colab reasons (`google.colab._kernel` import, missing `colab_kernel_launcher`, `-v 3.12`, etc.). Saying “nothing that would stop the runtime from reconnecting has been written, so this session is still usable” and “re-run with `-v 3.11`” is then wrong: `/usr/local` is already replaced. Re-running in the same runtime is the recovery path the ERR trap tells people not to use.

Please tell users to **Runtime > Disconnect and delete runtime** on any failure after step 3, same as the ERR trap. Shipping 7.0.0 on PyPI does not fix this message.

